PR #3, opened 2021-04-14
Added give functionality

Trey2k:
Created a simple give button, checks for the closest player within 3 units. Not the best at front end so may need some touch up for the margins. But it seems to work. If something needs changing let me know


Trey2k:
One sec, need to add a check for the amount of items.

Trey2k:
Okay should be all good now

PR #13, opened 2021-04-17
custom vehicle trunk

sahtiyan:


thelindat:
Not including per-vehicle weights in the base resource.

PR #18, opened 2021-04-17
Bug fix

Trey2k:
Fixed bug where two stacks with no meta data cannot be stacked


thelindat:
It'll do for now, but need to find out why items are losing their metadata (I assume it's only empty tables)

Trey2k:
I ran into the issue specifically when giving items to a player with a script with no meta data.

PR #47, opened 2021-05-06
global weight check

ElusionPDX:
checks if you can carry any item when given it


thelindat:
I'm not merging this in; the carry check is separate for a reason.

Issue #33, opened 2021-04-26
Text behind Image on Hotbar


Leah-UK:
Not an important / gamebreaking bug.

![image](https://user-images.githubusercontent.com/20465720/116080336-d68d0680-a690-11eb-8b78-154d62d5c323.png)

thelindat:
Can fix it, but images should be the same dimensions as included items so it doesn't clip with text at all.

PR #52, opened 2021-05-08
Limit draggable item to inventory bounds

PhotoshopLOL:
- set draggable containment to 'parent' instead of 'body'


PhotoshopLOL:
also added fix for #33

thelindat:
It'll do for now.

PR #96, opened 2021-06-03
Changes made to using item

Pezbian:
As discussed in Discord, changed use item functionality to allow for checks before using item if required and the ability to abort using the item if accidentally hit on hotbar for example. Feel free to change this up how you see fit however using items like this allows more flexibility when using the inventory. Also a couple of small QOL stuff with Progbar.

Changes:
- Modified to have a check event prior to progbar triggering. If the item doesnt have or need an event check (burger for example), it just proceeds to trigger progbar with no check.
- Modified to trigger event after progbar is complete
- Modified progbar to allow more customization (labels, ability to cancel).

Example used with Bandage as I did not want to create an item you didn't already have. Event 'bandageCheck' checks if the players heal is not max (200 HP). If it isn't, callback true and proceeds with progbar which has the ability to cancel if accidentally used. Once progbar is completed, it triggers to bandage event to heal the player.

### DIFF
--- a/client/items.lua
+++ b/client/items.lua
@@ -1,42 +1,37 @@
-AddEventHandler('linden_inventory:burger', function(item, wait, cb)
-	cb(true)
-	Citizen.Wait(wait)
+AddEventHandler('linden_inventory:burger', function(item)
 	TriggerEvent('mythic_notify:client:SendAlert', {type = 'inform', text = 'You ate a delicious burger', length = 2500})
 end)
 
-AddEventHandler('linden_inventory:water', function(item, wait, cb)
-	cb(true)
-	Citizen.Wait(wait)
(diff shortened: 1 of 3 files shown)


thelindat:
I removed the double-events stuff and added callbacks for a reason. I'll add the ability to cancel item usage in a different way, but right now I am focused on bugs.

Pezbian:
No problem. Based on how our server works, double events were far more suitable then the callback method especially when calling from other scripts. You would have one event call and finish > progbar > second event and was better resource wise then having one event just sit and wait for the progbar to do its thing. I understand you are looking at bugs, but at the same time the use item method was changed with minimal/no bugs so I was trying to help you write a better process.

PR #98, opened 2021-06-03
Enhancement/trigger event on item purchase

JDiMella:
Functionality to (optionally) trigger a server event when an item is purchased.

see issue: https://github.com/thelindat/linden_inventory/issues/97

### DIFF
--- a/server/main.lua
+++ b/server/main.lua
@@ -444,8 +444,18 @@ AddEventHandler('linden_inventory:buyItem', function(info)
 					else
 						removeInventoryItem(xPlayer, item.name, data.price)
 					end
-					addInventoryItem(xPlayer, data.name, count, data.metadata, false)
-					if Config.Logs then exports.linden_logs:log(xPlayer, false, ('bought %sx %s from %s for %s'):format(ESX.Math.GroupDigits(count), data.label, shopName, cost), 'items') end
+					-- if an onpurchase server event is defined for this item, trigger it and provide information about the Player, the Item an the Shop it's being purchased from
+					if Items[data.name].server ~= nil and Items[data.name].server.onpurchase ~= nil then
+						TriggerEvent(Items[data.name].server.onpurchase, xPlayer, Items[data.name], data, Config.Shops[location], function(result, metadata)
+							if result then
+								if metadata == nil then metadata = data.metadata end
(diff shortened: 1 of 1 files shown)


thelindat:
What's the intended use of triggering an event when buying an item

JDiMella:
more info in the issue - but here are a couple of use cases we have used this logic on our server:
* we had an external script keep track of the number of times a particular item is bought server wide (from any store). Upon hitting a limit, we want to prevent further sales of the item until restart
* we had custom ID cards being sold that upon purchase, would store the purchasing players information in the metadata of the item. So whenever the item was used, it would render HTML to nearby players based on the information in the ID card's metadata (allowing players to use other players ID card)
* We have some stores sell illegal items, so upon purchase, our event would fire and would notify police (based on RNG, and a few other factors like number players around etc) 
* When we use some of our items, it would apply a certain effect to the player based upon a value in the metadata. This value in the metadata is something we wanted to generate upon the purchase of the item based on external factors (the store it's bought from, time of day etc.)

JDiMella:
happy to chat in discord if you wish to discuss - Josh#2249

thelindat:
I'll keep this open for now - I have some planned changes for items that'll mess with it anyway, and giving people new tools to play with just to break it in an update isn't ideal.

thelindat:
Please take a look at the [dev branch](https://github.com/thelindat/linden_inventory/tree/dev) and submit any changes (or feedback) you would like there, instead. Example:

```lua
	['mustard'] = {
		label = 'Mustard',
		weight = 500,
		stack = true,
		close = true,
		client = {
			status = { hunger = 25000, thirst = 25000 },
			anim = { dict = 'mp_player_intdrink', clip = 'loop_bottle' },
			prop = { model = 'prop_food_mustard', pos = { x = 0.01, y = 0.0, z = -0.07 }, rot = { x = 1.0, y = -1.0, z = -1.5 } },
			usetime = 2500,
			event = true,
		},
		server = {
			onbuy = 'linden_inventory:mustardcheck',
			onadd = 'linden_inventory:mustardmetadata'
		}
	},
```

JDiMella:
updated the PR based upon latest dev - let me know what you think 😄

thelindat:
Merged and tweaked.

Would appreciate if you gave me an example usage so I can throw a few events in and update docs.

Issue #99, opened 2021-06-04
Ability to prevent specific items from being used in a vehicle


JDiMella:
**Is your feature request related to a problem? Please describe.**
It would be good to have the ability to define if an item should/shouldn't be permitted to be used whilst the player is in a vehicle.

**Describe the solution you'd like**
(optionally) prevent an item from being used if the player is in a vehicle. Can be achieved as follows:
* add an optional key to an item's config: `canUseInVehicle` that is either true or false. (or if key not present, can assume is true)
* when an item is used, check if a player is in a vehicle, and check the config for that item to see if `canUseInVehicle` is defined and set to `false`. If so, prevent the item from being used. Otherwise, assume the item can be used in a vehicle and proceed.

**Describe alternatives you've considered**
You could implement this check in the individual event for each item, but that would be a lot of repetition and isn't very DRY

**Additional context**
I'll throw up a PR with a suggested solution shortly.

JDiMella:
pr: https://github.com/thelindat/linden_inventory/pull/100

PR #100, opened 2021-06-04
Add check to optionally prevent items being used if in a vehicle

JDiMella:
Functionality to (optionally) prevent an item being used whilst in a vehicle.

see issue: #99


thelindat:
Please take a look at the [dev branch](https://github.com/thelindat/linden_inventory/tree/dev) and submit any changes you would like there, instead.

JDiMella:
I've re-based on-top of dev & updated the PR to merge into dev instead of main 😄

JDiMella:
that's been updated to use the new item structure 👍

PR #101, opened 2021-06-05
Update for usable parachute item

jetpnix:


thelindat:
Parachutes need to use a proper event rather than being used like a weapon.

PR #102, opened 2021-06-06
Create pl.lua

TeksuSiK:


TeksuSiK:
Also I suggest to make rest of buttons, lores etc. translatable, because there are currently hardcoded

PR #107, opened 2021-06-08
Set target player unarmed

FreakOutModz:
Player will be unarmed when inventory is open by another player with /steal or by police. To be test before push.


thelindat:
If this bug is indeed occurring, and I will say I've already patched and tested disarming players by taking their weapons, then there is a proper method for disarming.

PR #108, opened 2021-06-08
Idea propose by player for admin to open inv

FreakOutModz:
Admin will be able to open inventory of player like /steal, but with /adminopen


thelindat:
`xPlayer.getGroup()` is not a valid client-side function, and any method for doing so needs to be safe from exploitation - I already have something in mind but it is not a priority.

PR #109, opened 2021-06-08
Version check for nothing

FreakOutModz:
You have the version: '1.6.4' so you don't need this. You should also start to think about getting mysql instead of ghmattimysql.


thelindat:
Version check is for checking the most recent version on github and informing you that you do not have a matching version.
As for the supported resource for queries, I like ghmattimysql.

PR #111, opened 2021-06-08
Update main.lua

FreakOutModz:
Uptade little things in main


thelindat:
Not sure what you're trying to do here - the drawtext is no more optimised than the current (and missing some stuff); ESX.PlayerData.ped is a stored value of your current ped so no need to call the native which is slightly slower.

PR #118, opened 2021-06-10
Add progressbars for weight display

OfficialNoms:


OfficialNoms:
I don't know why it does this dumb shit I need to refork, but yeah

PR #123, opened 2021-06-14
Add bt-target Functions

Leah-UK:
Not sure if this is something you wan't to implement into the core code. But this is a good base for implementing bt-target directly into the inventory (since people seem to like doing that).

**Added:**
- Config for enabling / disabling.
- Width / Length of bt-target in shops.lua
- Opening of shop through bt-target in client/main.lua


Leah-UK:
Just realised I pushed a lot of other shit from my version. Can remove it if this is something you actually want to merge :)

PR #142, opened 2021-06-21
Fixes for weapon switching

d3rp-jsx:
This should fix hotbar weapons switching definitely


thelindat:
That code is for holstering - if the weapon being used has the same serial then holster. Melee weapons, thrown weapons, and misc weapons do not have serials, so it checks for name.

d3rp-jsx:
@thelindat changing the 'or' to an 'and' will result in checking for both name and serial so you fix unintended holster while:
- switching from a weapon to the same weapon with different serial number
- switching between weapons without serials (where serial check is both nil so true, then holster)

thelindat:
This prevents you from putting melee weapons away entirely

d3rp-jsx:
> This prevents you from putting melee weapons away entirely

Nope because serial will always be nil = nil so true, so the name only must be true.

thelindat:
image

d3rp-jsx:
https://user-images.githubusercontent.com/15928886/123610716-227b3980-d801-11eb-96cc-d8e8f74d5b15.mp4

@thelindat there, it's working as expected for me

d3rp-jsx:
@thelindat I'm running my server with this commit for weapon switches for days (even updating the inventory itself) and it's working perfectly, you should review this.

PR #162, opened 2021-07-01
Update LICENSE

thelindat:
Due to the restrictive nature of Creative Common licenses that do not make them particularly appropriate for software, I am pushing to change the license to something that grants the most permissions to its users and contributors.

I initially accepted the change of license to the current one because I was completely fine with the terms as I understood them, however it does force some limitations on what I can do with the code in regards to other third-party resources without breaking the license terms of both.

As there are still large parts of the code (notably the javascript) which belong to Hasan, this is pending express permission.


hsnnnnn:
Changing the license is fine, you can use it!

PR #201, opened 2021-07-29
Fixed wrong name for slots in bag inventory

dolutattoo:
Just changed `item.metadata.slot` to `item.metadata.slots`

### DIFF
--- a/client/main.lua
+++ b/client/main.lua
@@ -1107,7 +1107,7 @@ RegisterNetEvent('linden_inventory:useItem')
 AddEventHandler('linden_inventory:useItem', function(item)
 	if item.metadata.bag and not currentInventory then
 		invOpen = false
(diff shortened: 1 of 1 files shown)


thelindat:
This won't really fix anything afaik? Nothing actually assigns metadata.slot and this is the only place it gets checked, but bags themselves are only in a very basic state.

The idea is that you create your own bags and assign metadata.maxWeight and metadata.slot.

Though metadata.size might be a better name.

dolutattoo:
That's not really a fix yes, but I think it makes way more sense to call all slot limits the same way.
I made this change cause for stashes, the value is called `slots`, and people will start to configure their inventory trough stashes before bags.
So if people try to use the bag feature, they will probably have the same problem as me (not seeing that bags slot limit was called `slot` and not `slots`)

Thats more a suggestion, sorry if that is not appropriate in a pull request, im pretty new to git stuff

PR #253, opened 2021-09-18
fix(refactored): sql query and config.

coookie1010:
There is a open parenthesis without closing in sql query and adjusted the
Interval literals.

http://docs.oracle.com/cd/B19306_01/server.102/b14200/sql_elements003.htm

### DIFF
--- a/config.lua
+++ b/config.lua
@@ -11,7 +11,7 @@ if ox.server then
 	Config.CheckVersion = false
 
 	-- When should unused inventories be wiped
-	Config.DBCleanup = '6 MONTH'
+	Config.DBCleanup = '6' -- Months
 	
 	-- Enable integrated logging
 	Config.Logs = false
--- a/modules/items/server.lua
+++ b/modules/items/server.lua
@@ -72,7 +72,7 @@ CreateThread(function()
 	end
 	Wait(2000)
 	TriggerEvent('ox_inventory:itemList', Items)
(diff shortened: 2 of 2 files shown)


modules/items/server.lua
```diff
@@ -72,7 +72,7 @@ CreateThread(function()
 	end
 	Wait(2000)
 	TriggerEvent('ox_inventory:itemList', Items)
-	if Config.DBCleanup then exports.oxmysql:executeSync('DELETE FROM ox_inventory WHERE lastupdated < (NOW() - INTERVAL ? OR data = "[]"', {Config.DBCleanup}) end
+	if Config.DBCleanup then exports.oxmysql:executeSync('DELETE FROM ox_inventory WHERE lastupdated < (NOW() - INTERVAL ? MONTH OR data = "[]")', {Config.DBCleanup}) end
```
dunak-debug, line 75: You cannot put `NOW() - INTERVAL X OR data = "[]"` inside parantheses. This is not valid query.

We are checking **interval or empty data.** 
```sql
WHERE lastupdated < (NOW() - INTERVAL ?) OR data = "[]"
```

config.lua
```diff
@@ -11,7 +11,7 @@ if ox.server then
 	Config.CheckVersion = false
 
 	-- When should unused inventories be wiped
-	Config.DBCleanup = '6 MONTH'
+	Config.DBCleanup = '6' -- Months
```
dunak-debug, line 14: What if user wants to clean it up after `X DAY` ? Also it is more readable when there are month inside config imo.

dunak-debug (changes requested):

thelindat:
It's already fixed on my end, but my git has been messed up. I'd rather not force the cleanup setting to use `MONTH`.

PR #265, opened 2021-10-05
fix(shops): Remove durability bar on shop items

Skryptific:
Sometimes when you purchase an weapon and use it, it would show a durability bar on the item in the shop inventory.


LukeWasTakenn:
First commit looks good. @thelindat needs to take a look at the second one though

thelindat:
I'll look at swapitems some more after this.

PR #267, opened 2021-10-08
Add locales

Skryptific:
Hi, I noticed that there are a few locales missing so I added them and added dutch translations while I was at it.


LukeWasTakenn:
LGTM, thank you

PR #272, opened 2021-10-10
feat(locales): French locale

davidenco2000:
Add french language to inventory


LukeWasTakenn:
Just so you know the locales aren't final, they are probably going to need adding onto moving towards the end of the development of the 2.0 update.

PR #319, opened 2021-11-04
fix(server): Add count to creating drops

Skryptific:
Referring to [this issue](https://trello.com/c/wcjYaqMD/65-creating-drops-ignores-count) on the Trello page.


thelindat:
Applied a small fix to prevent exploitation of the event. Thanks for the contribution.

PR #321, opened 2021-11-04
feat(shops): Add qtarget support to shop peds

Skryptific:
Hi,

I've added this for my own server and thought it would be cool to contribute. It allows users to configure and spawn peds (very minimal configuration though) for each individual shop instead of zones when using qtarget. You can set `ped` and `task` under `targets`, see the examples I have added in data/shops or this commit 263b6fb0f909406a755b3f7ed00db57e02a65a9f. Task is optional, but would be cool to use in some cases.

Regular target as normal.
target_1

Newly supported target.
target_2

Hope this is useful. :)

### DIFF
--- a/data/shops.lua
+++ b/data/shops.lua
@@ -148,9 +148,11 @@ return {
 			{ name = 'ammo-rifle', price = 1000 },
 			{ name = 'ammo-rifle2', price = 1000 }
 		}, locations = {
-			vec3(309.09, -913.75, 56.46)
+			vec3(210.55, -1989.15, 19.71),
+			vec3(336.08, -2021.74, 22.35)
 		}, targets = {
-
+			{ loc = vec3(210.55, -1989.15, 19.71), length = 0.5, width = 1.5, heading = 320.0, minZ = 18.5, maxZ = 21.25, distance = 3.0 },
+			{ loc = vec3(336.08, -2021.74, 21.35), heading = 140.0, distance = 1.5, ped = `mp_g_m_pros_01`, task = 'WORLD_HUMAN_AA_SMOKE' }
 		}
 	}
 }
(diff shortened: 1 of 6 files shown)


LukeWasTakenn:
Having peds wasn't really an idea with shops, although can be left to @thelindat to decide whether he wants this.

Also I see you are creating and spawning peds but I don't see any code for removing them. There would need to be some kind of distance check to create and remove peds, otherwise you will hit the game pool cap for peds quite fast. Although I feel like doing distance checks instead of PolyZone (which we don't use) isPointInside checks kinda defeats half the purpose of qtarget?

As I've said, up to @thelindat.

Skryptific:
> 
> 
> Having peds wasn't really an idea with shops, although can be left to @thelindat to decide whether he wants this.
> 
> Also I see you are creating and spawning peds but I don't see any code for removing them. There would need to be some kind of distance check to create and remove peds, otherwise you will hit the game pool cap for peds quite fast. Although I feel like doing distance checks instead of PolyZone (which we don't use) isPointInside checks kinda defeats half the purpose of qtarget?
> 
> We'd also have to add peds to _all_ of the shops otherwise it might just feel weird where at some places there are peds and at others there aren't.
> 
> As I've said, up to @thelindat.

Depending on what Linden wants, I will write something to handle the peds. 

I added the support so people can choose whatever they want to use and when, I can imagine that someone wants the cash register to be the target for normal shops and hidden shops or whatever to be a ped.

thelindat:
This is neat and all, however the project already has a lot of scope creep occurring and extra features will require additional documentation (which I'm already slacking on); bt-target and later qtarget support was added by request and community submission, but it's never been something I've actively worked on integrating.

This could be acceptable in the future, however.

PR #324, opened 2021-11-05
feat(inventory): Add view inv admin command

Skryptific:
Allows admins to view the inventory of a player by id, added a check if user is an admin or else the inventory gets closed when not nearby.

Issue reference: https://trello.com/c/9RCP0LOa/7-view-target-inventory

### DIFF
--- a/client.lua
+++ b/client.lua
@@ -421,7 +421,7 @@ RegisterNetEvent('ox_inventory:setPlayerInventory', function(drops, inventory, w
 						local id = GetPlayerFromServerId(currentInventory.id)
 						local ped = GetPlayerPed(id)
 						local pedCoords = GetEntityCoords(ped)
-						if not id or #(playerCoords - pedCoords) > 1.8 or not (ESX.PlayerData.job.name == 'police' or CanOpenTarget(ped)) then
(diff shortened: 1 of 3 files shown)


client.lua
```diff
@@ -421,7 +421,7 @@ RegisterNetEvent('ox_inventory:setPlayerInventory', function(drops, inventory, w
 						local id = GetPlayerFromServerId(currentInventory.id)
 						local ped = GetPlayerPed(id)
 						local pedCoords = GetEntityCoords(ped)
-						if not id or #(playerCoords - pedCoords) > 1.8 or not (ESX.PlayerData.job.name == 'police' or CanOpenTarget(ped)) then
+						if not id or (#(playerCoords - pedCoords) > 1.8 or not (ESX.PlayerData.job.name == 'police' or CanOpenTarget(ped)) and ESX.PlayerData.group ~= 'admin') then
```
LukeWasTakenn, line 424: This would still do the distance check for admins? we kinda want admins to be able to access it globally

thelindat:
I'm planning to add the functionality for this in a different way with some checks for security purposes; storing the player group in ESX.PlayerData with a callback feels a little hacky and doesn't provide protection against injection.

Skryptific:
I've done some changes just in case you would like to look at it. https://github.com/Skryptific/ox_inventory/commit/dc05d556ed1a48274d8e32046913e24fb4817fd1

PR #330, opened 2021-11-09
Player IDs option for giving item when nearby multiple players

jetpnix:
Forgot to screenshot, but in general it shows an option when there are 2 or more players nearby. If there is only one, it will still use the default closest player.

Please do check! First time transitioning to typescript from javascript


web/src/components/inventory/ShowNearby.tsx
```diff
@@ -0,0 +1,67 @@
+import { useState } from 'react';
+import useNuiEvent from '../../hooks/useNuiEvent';
+import { giveTo } from '../../dnd/onGive';
+import { useExitListener } from '../../hooks/useExitListener';
+
+interface Player {
+  player: number;
+}
```
LukeWasTakenn, line 8: Why make an interface with just one number value in it?

web/src/components/inventory/ShowNearby.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+const ShowNearby: React.FC<{ isInventoryOpen: boolean }> = ({ isInventoryOpen }) => {
+  const [nearbyVisible, setNearbyVisible] = useState(false);
+  const [nearbyPlayers, setNearbyPlayers] = useState([]);
+  const [currentSlot, setSlot] = useState(0);
+  const [currentAmount, setAmount] = useState(0);
+
+  useNuiEvent('showNearby', (data) => {
+    if (nearbyVisible) {
```
LukeWasTakenn, line 17: You should be closing inventory on dialog open, looks awkward when both are open at the same time.

web/src/components/inventory/ShowNearby.tsx
jetpnix: Ah sorry bout that. Again this is the very first time i've tried typescript, do improve it if i was wrong.

web/src/components/inventory/ShowNearby.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+interface Player {
+  player: number;
+}
+
+const ShowNearby: React.FC<{ isInventoryOpen: boolean }> = ({ isInventoryOpen }) => {
+  const [nearbyVisible, setNearbyVisible] = useState(false);
+  const [nearbyPlayers, setNearbyPlayers] = useState([]);
+  const [currentSlot, setSlot] = useState(0);
+  const [currentAmount, setAmount] = useState(0);
```
LukeWasTakenn, line 14: These can probably be a state object since you are setting them all at once, no need for 3 separate states.

web/src/components/inventory/ShowNearby.tsx
```diff
@@ -0,0 +1,67 @@
...
+  };
+
+  return (
+    <>
+      {isInventoryOpen && (
+        <div className="center-wrapper">
+          <div className="nearby-box" style={{ visibility: nearbyVisible ? 'visible' : 'hidden' }}>
+            {nearbyPlayers.map((player: Player) => {
+              return (
+                <div className="player-id" onClick={() => triggerGiveTo(player.player)}>
```
LukeWasTakenn, line 55: Not really an issue but `player.player` should avoid having properties the same name as the object.

LukeWasTakenn (commented):
Overall code looks *ok*.

Would need @thelindat to go over the Lua side but even that to me looks *ok*.
 Depends on him whether he even wants this functionality or not.

web/src/components/inventory/ShowNearby.tsx
LukeWasTakenn: nearbyVisible is in the block but that's fine as is, it can be it's own state.

dunak-debug:
Ill take a look at nui side in a while

LukeWasTakenn:
> Ill take a look at nui side in a while

I went through it, looked fine to me but sure double check

web/src/components/inventory/ShowNearby.tsx
jetpnix: Ok fixed it, it should only contain player ids now, not object with player key and player id value.

web/src/components/inventory/ShowNearby.tsx
jetpnix: I'm not sure how to do it without messing up the structure of the codes. I'm thinking of doing this with players able to close the dialog when clicking outside the dialog. Not sure where to place the code, maybe in the hooks? and honestly not too confident if mine will be a good code haha.

web/src/components/inventory/ShowNearby.tsx
jetpnix: Done made changed it as an object in the latest commit

web/src/components/inventory/ShowNearby.tsx
jetpnix: Fixed it, should now be a number instead of an object

jetpnix:
image
Here is the preview of the dialog when open. @LukeWasTakenn how did you test the nui? Just asking for my next testing or edit 😄

dunak-debug:
Surely this needs to have closed inv or at least make opacity 100% for give dialog. It looks weird now.. also @thelindat have to say if he wants this functionality

dunak-debug:
Another idea would be to open context menu with player ids

LukeWasTakenn:
> @LukeWasTakenn how did you test the nui?

What do you mean exactly?

thelindat:
Appearance isn't great and I've said before it's not really something I want.

For interaction stuff like this, it makes more sense to utilise qtarget (or whatever you prefer) - in which case we would need to start the give function and await player selection.

jetpnix:
> 
> 
> Appearance isn't great and I've said before it's not really something I want.
> 
> For interaction stuff like this, it makes more sense to utilise qtarget (or whatever you prefer) - in which case we would need to start the give function and await player selection.

Right, the appearance i can work on it. For qtarget seems viable but i feel creating a context menu like dunak said would be simpler. Well i'll use this for now and wait for your method of choosing a player to give when there are multiple players nearby.



> 
> 
> > @LukeWasTakenn how did you test the nui?
> 
> What do you mean exactly?

Umm did you test the UI? using `yarn start` or ingame? Since i'm unable to test the full features in `yarn start` or its just that im dumb.

LukeWasTakenn:
> Umm did you test the UI? using `yarn start` or ingame? Since i'm unable to test the full features in `yarn start` or its just that im dumb.

In game by building it or using `yarn start:game`. We haven't been doing development in the browser for quite some time now.

Senlar:
I agree with thelindat, the better option here is qtarget. Its often awkward when you are in a large group of people to have to make sure you press the button to see the players ID then remember what it was and select it from the give command. Much more precision and ease of use would come from the qtarget option.

However, I think this is a nice option for people who do not use qtarget.

thelindat:
qtarget support is there to add extra features and make it easier to utilise the resource, especially for cases like this. I believe the best option for improving standard use would just be utilising a raycast to get the player in front of you (or in the seat besides you, for vehicles).

I never wanted give in the first place, but it was PR'd so I accepted it; adding a give window on top of that isn't something I'm likely to change my mind on.

thelindat:
Closing for now; if you have suggestions or improvements feel free to submit a new PR.

PR #344, opened 2021-11-21
Limit the quantity for an item (Optional)

dolutattoo:
I think that's a must have to be able to prevent a player to carry more than X quantity of an item, even if the weight capacity is ok.
It's probably not the best way to do it at all, I'm a complete beginner, but at least you get the idea.

For example, you don't want to see a player with 15 heist bags in his inventory.
Or if you want to physically attach a chicken to the hand of a player when he just catched one from the farm.

PS: This was tested in `shops`, `trunks`, `gloveboxes`, and `give` feature with 3 players on the server. No issue for us, except when you `/steal` someone, didnt found where to put the limit (because of time, I work f*ing slowly ^^)

### DIFF
--- a/data/items.lua
+++ b/data/items.lua
@@ -130,6 +130,7 @@ return {
 	['water'] = {
 		label = 'Water',
 		weight = 500,
+		limit = 15,
 		client = {
 			status = { thirst = 200000 },
 			anim = { dict = 'mp_player_intdrink', clip = 'loop_bottle' },
--- a/data/items.lua
+++ b/data/items.lua
@@ -29,6 +29,7 @@ return {
 		label = 'Test Burger',
 		weight = 220,
 		degrade = 60,
+		limit = 3,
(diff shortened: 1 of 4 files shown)


thelindat:
While I don't see the point of limit in many cases, I'd rather see it implemented well rather than hacky methods (and not have to do it myself). Complete the requested changes before I test and do my own cleanup.

thelindat (changes requested):

dolutattoo:
I cannot use GetItemSlots inside modules files

thelindat:
> I cannot use GetItemSlots in modules/*.lua
> 
> It require to edit: `local GetItemSlots = function(inv, item, metadata)` by this: `M.GetItemSlots = function(inv, item, metadata)`

Yes, that's fine.

dolutattoo:
All requested changes are done

PR #347, opened 2021-11-24
Ox_INVENTORY must be started after QTARGET if this is TRUE

Velkostatkar:
chaged in cfg


LukeWasTakenn:
It's already in the docs, doesn't need to be in the config file, also it makes sense that a resource with exports needs to be started before.

PR #354, opened 2021-11-30
Fix stack bug on giveitem

tony-stark-17:


thelindat:
This is more likely to be from a change to the table.matches functionality at some point; should be a one-liner.

thelindat:
Resolved with https://github.com/overextended/ox_inventory/commit/16216b98a8b1bcc1d4011469abaf8e154d0391ca

PR #355, opened 2021-11-30
feat: Ability to define slots and maxWeight for a specific vehicle model

vWernay:


thelindat:
Going to merge for now, needs some refactoring overtime.

PR #368, opened 2021-12-10
refactor: Unify indentation and formatting

antond15:
And add some missing locales


thelindat:
Resolve conflict, also some stuff is spaces by design (fxmanifest for example).

PR #376, opened 2021-12-13
Add check in CanSwapItem for inventory ID

baguscodestudio:
Check for inventory ID instead of inventory object or table


thelindat:
Mildly upset nobody reported this bug 😢

PR #388, opened 2021-12-17
fix(inventory): Server

DogsGoBark:
- Fixed `clearinv` to use exports.


thelindat:
That's not a fix or a proper commit message

PR #394, opened 2021-12-20
fixed double adding component in table when mounts it

hAKIZOR:
when set up a component on a weapon it goes 2 times in the table of the components... so with this issue is easy duplicate components.

### DIFF
--- a/modules/inventory/server.lua
+++ b/modules/inventory/server.lua
@@ -783,6 +783,19 @@ RegisterServerEvent('ox_inventory:giveItem', function(slot, target, count)
 	end
 end)
 
+local function hasValue( tbl, str )
+    local f = false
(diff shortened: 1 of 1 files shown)


thelindat:
There should be something in place preventing you from using an attachment if the component exists already, so if there's a problem of doubling up it should be resolved elsewhere.

hAKIZOR:
> 

but at row 595 in modules/inventory/server.lua there is an adding in table before removing the component as item from inventory... and this component will be added again in the event "updateWeapon"

LukeWasTakenn:
I'm pretty sure when I wrote this I was doing a check to see if a the component already exists on the weapon...?

thelindat:
Resolved.

PR #396, opened 2021-12-21
Disarm player, when clearing inventory.

icecubepiso:


thelindat:
Implemented fix on client-side.

PR #421, opened 2022-01-01
fix(server): Use "evidence" as name for evidence

antond15:
`/clearevidence` command deletes inventories with `evidence` prefix, not `police`.

Maybe it's better to use `police`, because somebody can already have database entires with this prefix, but I chose `evidence` as it makes more sense for future references.
Swap it if you want.


thelindat:
This is actually the intended prefix 😥

PR #424, opened 2022-01-01
feat(items): Build item use functions based off client.notification

Vicken52:
Did this to make life easier and to trim down on duplicate code.

Not always needed, if you want to build the function yourself, but there if you're lazy

### DIFF
--- a/data/items.lua
+++ b/data/items.lua
@@ -39,6 +39,7 @@ return {
 			anim = { dict = 'mp_player_inteat@burger', clip = 'mp_player_int_eat_burger_fp' },
 			prop = { model = `prop_cs_burger_01`, pos = vec3(0.02, 0.02, -0.02), rot = vec3(0.0, 0.0, 0.0) },
 			usetime = 2500,
+			notification = 'You ate a delicious %s'
 		},
 	},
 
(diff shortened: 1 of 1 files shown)


modules/items/client.lua
```diff
@@ -22,6 +22,17 @@ local ox_inventory = exports[ox.resource]
 -----------------------------------------------------------------------------------------------
 -- Clientside item use functions
 -----------------------------------------------------------------------------------------------
+for key, value in pairs(Items) do
```
thelindat, line 25: 🔕 This won't do, though I do have a good solution for it. Remove this change and squash, then I'll add another change on top for handling notifications.

The notification stuff should be an improvement for all items to be able to utilise, rather than reducing the number of items that need to be explicitly registered. There should already be something in place that sets these items as usable and will fallback to a generic use event without a callback (I'll double check).

thelindat (changes requested):

modules/items/client.lua
Vicken52: 👍 Reverted everything in `items/client.lua` back to normal

thelindat:
I'll have a look at everything after dinner, thanks for the contributions 👍

PR #425, opened 2022-01-01
feat(items): Reusable references for animations and props

Vicken52:
More changes to help reduce duplicate code and make adjusting animations and props easier if needed across multiple items.


thelindat:
💸

PR #433, opened 2022-01-05
Almost 100% Hungarian translate

Rav3n95:


dunak-debug:
Almost 100%?

Rav3n95:
> 

Few lines are missing, also Hungarian language are really complicated compare to English. Propably will have to correct few words here and there.

dunak-debug:
> Few lines are missing, also Hungarian language are really complicated compare to English. Propably will have to correct few words here and there.

Alright just make it draft until it’s done

thelindat:
@Rav3n95 is this ready?

PR #442, opened 2022-01-06
Notify if use component and not wearing weapon

dolutattoo:
There is a locale who was waiting for it so let's use it :)


LukeWasTakenn:
Thank you Dolu

PR #443, opened 2022-01-09
Disable Rolling while reloading to prevent bug

tony-stark-17:


thelindat:
Wasn't aware there _was a bug_ since there isn't an issue. From testing, I don't see it as much of an issue (some ammo sits in reserve and will be properly loaded after re-equipping or reloading).

This can be done a lot cleaner without the added interval, anyway.

PR #451, opened 2022-01-11
Fix inventory not closing after item usage

baguscodestudio:
This will now close inventory for item with properties of close = true


baguscodestudio:
Need to use exports so it will register as an item. Bad PR shucks
https://github.com/overextended/ox_inventory/blob/main/data/items.lua#L2-L17
https://github.com/overextended/ox_inventory_examples/blob/main/client.lua#L18-L26

PR #454, opened 2022-01-12
Better durability

GaluzaCZ:
Mby better weapons durability


thelindat:
While I agree the current values set are excessive, I'd like some feedback from others if these values feel more balanced.

og23rd:
Hes right

antond15:
I've been using very similar values for few days and it's quite well balanced.

PR #456, opened 2022-01-14
fix(exports): use ox.resource resource name variable instead of static ox_inventory in export calls

Snowthisway:
Only use ox.resource (defined in config, default GetCurrentResourceName()) for appropriate export calls instead of static "ox_inventory".


LukeWasTakenn:
Is this so people can just rename the resource and claim it as theirs more easily?

Snowthisway:
> Is this so people can just rename the resource and claim it as theirs more easily?

This change is consistent with what already exists in modules/items/client.lua and makes the config variable actually do what its supposed to.

LukeWasTakenn:
Right, I see. I'll leave it up to @thelindat to decide since he implemented the `ox.resource` variable in the first place.

PR #466, opened 2022-01-16
fix(client): Pass item count to remove function

antond15:
And disable phone only if count is 0.

### DIFF
--- a/client.lua
+++ b/client.lua
@@ -609,7 +609,7 @@ local function updateInventory(items, weight)
 				data.count += v.count
 			else
 				local data = Items[item.name]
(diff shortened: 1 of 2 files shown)


thelindat:
That function, from testing, should only trigger when the item is removed entirely; or did I miss a scenario?

antond15:
> That function, from testing, should only trigger when the item is removed entirely; or did I miss a scenario?

If you had 2 phones in inventory, it disabled the phone when you moved one item away.

antond15:
Okay, while testing this, I found another problem. Created an issue #468.

PR #478, opened 2022-01-19
Better notification when trying to carry too much

dolutattoo:
I made those changes on my server and this is very easier for player to understand what is wrong when trying to swap, buy, give or rob items.

This is also probably better for people because they won't have to handle the notification themself when using `Inventory.CanCarryItem`.

I added locales for each languages but only french is translated.

image

### DIFF
--- a/data/locales/cs.lua
+++ b/data/locales/cs.lua
@@ -50,6 +50,8 @@ return {
 	['cannot_perform'] = "Tuto akci nelze provést",
 	['cannot_carry'] = "Tolik toho neunesete",
 	['cannot_carry_other'] = "Druhý inventář toho tolik neunese",
+	['cannot_carry_limit'] = "You only can carry %s %s maximum",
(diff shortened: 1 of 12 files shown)


thelindat:
Can you squash these commits with a message like `refactor: carry limit notifications` or something.

dolutattoo:
> Can you squash these commits with a message like `refactor: carry limit notifications` or something.

I'm not sure that I can make PR with Fork, so I'm always doing that manually on github.com ^^
How can I squash them now?

thelindat:
> > Can you squash these commits with a message like `refactor: carry limit notifications` or something.
> 
> I'm not sure that I can make PR with Fork, so I'm always doing that manually on github.com ^^ How can I squash them now?

I'll take care of it for now, but I recommend at least setting a better commit message in the future since it's easier to track down specific changes.

PR #494, opened 2022-01-22
feat(Logs) Robbery logged with OpenPlayerInventory item swaps

Senlar:
This adds logging for when police or civilians access another players inventory and move items in or out. You may want to tweak this before merging to fit with the new DataDog integration. I haven't done that yet as I was still using the discord logging.

### DIFF
--- a/server.lua
+++ b/server.lua
@@ -199,6 +199,7 @@ ServerCallback.Register('swapItems', function(source, data)
 			local sameInventory = fromInventory.id == toInventory.id or false
 			local container = (not sameInventory and playerInventory.containerSlot) and (fromInventory.type == 'container' and fromInventory or toInventory)
 			local containerItem = container and playerInventory.items[playerInventory.containerSlot]
+			local robbery = nil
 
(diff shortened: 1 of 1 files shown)


Senlar (commented):
I think this adjustment will make it work with the new Datadog format

thelindat:
This would be better with a tag, rather than modifying the message IMO.

`'swapSlots', fromInventory.owner or fromInventory.id, toInventory.owner or toInventory.id`
Those are the existing tags; adding a robbery tag would look cleaner. I don't think there's much need for distinction between robbery and police confiscation but 🤷

Senlar:
The difference for me is I want to be able to tell if police or non-police are abusing the function. We have RP rules against people just going around and "pocket wiping" people but our police need to frequency take certain items to put into evidence before sending a criminal off to jail or back onto the streets. So that is why the distinction is important to us.

When I first wrote this I didn't have any concept on how datadog logs worked but now I see the tags and I agree that is the better way to do it.

thelindat:
I'll look at more default tags in the future.

PR #496, opened 2022-01-23
Adding deleteinventory export

ffpetrozzi:


antond15 (changes requested):
This will only delete the inventory from database.
You have to remove it from `Inventories` table too.

thelindat:
what

PR #511, opened 2022-02-02
Create ba.lua

katotekii:
here you go luke


data/locales/ba.lua
```diff
@@ -0,0 +1,86 @@
...
+	['ui_removeattachments'] = "Skini Dodatke",
+	-- Tooltip
+	['ui_durability'] = "Izdržljivost",
+	['ui_ammo'] = "Metci",
+	['ui_serial'] = "Serijski Broj",
+	['ui_components'] = "Komponente",
+	['ui_tint'] = "Tinta",
+	-- Useful controls
+	['ui_usefulcontrols'] = "Korisne kontrole",
+	['ui_rmb'] = "Item Contex Menu",
```
LukeWasTakenn, line 16: Perhaps instead of this just do "Item kontekst meni"?

data/locales/ba.lua
```diff
@@ -0,0 +1,86 @@
...
+	['ui_ctrl_lmb'] = "Brzo premjestite hrpu predmeta u drugi inventar",
```
LukeWasTakenn, line 17: Here you use "predmeta" but in most cases you use "item" so should really decide between the two

I think "predmet" is probably a better translation all around

data/locales/ba.lua
```diff
@@ -0,0 +1,86 @@
...
+	['ui_shift_drag'] = "Razdvoji na pola iteme",
```
LukeWasTakenn, line 18: Should be "Razdvoji iteme na pola"

LukeWasTakenn (changes requested):

katotekii (commented):
here you go

LukeWasTakenn:
Hvala punoo

PR #520, opened 2022-02-04
Drop and component fix

antond15:
- Fixed not opening drop inventory. This was caused by changed value in `PlayerData.id`.
- Fixed extended clip attachment for Combat MG Mk II.
- Removed some old debug stuff.

### DIFF
--- a/client.lua
+++ b/client.lua
@@ -672,6 +672,7 @@ local function updateInventory(items, weight)
 	end
 
 	client.setPlayerData('inventory', PlayerData.inventory)
+	TriggerEvent('ox_inventory:updateInventory', changes)
 end
 
 RegisterNetEvent('ox_inventory:updateSlots', function(items, weights, count, removed)
@@ -817,6 +818,7 @@ RegisterNetEvent('ox_inventory:setPlayerInventory', function(currentDrops, inven
 	plyState:set('invBusy', false, false)
 	plyState:set('invOpen', false, false)
 	PlayerData.loaded = true
+	TriggerEvent('ox_inventory:updateInventory', PlayerData.inventory)
 
 	Utils.Notify({text = shared.locale('inventory_setup'), duration = 2500})
 	local Licenses = data 'licenses'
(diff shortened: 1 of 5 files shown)


client.lua
```diff
@@ -672,7 +672,6 @@ local function updateInventory(items, weight)
 	end
 
 	client.setPlayerData('inventory', PlayerData.inventory)
-	TriggerEvent('ox_inventory:updateInventory', changes)
```
antond15, line 675: I am not 100% sure if this is left from testing `add` and `remove` options or if it's something with a similar purpose as `ox_inventory:currentWeapon`, for example.

thelindat: This is for other resources to listen for changes; can be used for add/remove functionality and metadata changes.

client.lua
```diff
@@ -706,7 +705,7 @@ RegisterNetEvent('ox_inventory:createDrop', function(drop, data, owner, slot)
 		drops[drop] = data
 	end
 
-	if owner == PlayerData.id and invOpen and #(GetEntityCoords(PlayerData.ped) - data.coords) <= 1 then
+	if owner == GetPlayerServerId(PlayerData.id) and invOpen and #(GetEntityCoords(PlayerData.ped) - data.coords) <= 1 then
```
thelindat, line 708: Did I end up storing PlayerData.source? If not then add that during init and use here.

modules/items/server.lua
```diff
@@ -121,39 +121,6 @@ local itemFormat = [[
...
-		local Players = ESX.GetPlayers()
-		for i = 1, #Players do
-			local i = Players[i]
-			--if not IsPlayerAceAllowed(i, 'command.refresh') then
-				local inv, ped = Inventory(i), GetPlayerPed(i)
-				local hash, curWeapon = GetSelectedPedWeapon(ped)
-				if not ignore[hash] then
-					curWeapon = Utils.GetWeapon(hash)
-					if curWeapon then
-						local count = 0
```
thelindat, line 136: This is part of detecting weapon cheats that I need to go back and rewrite; old code is left behind for quick reference.

thelindat:
Would be best to drop the commits rather than recommitting what was removed

PR #525, opened 2022-02-06
feat(shops): Fixed price option

antond15:
Useful for some items like weapons, tickets, ID card etc.


thelindat:
I think I'd prefer removing random prices and calling it as a function for price instead.

antond15:
> I think I'd prefer removing random prices and calling it as a function for price instead.

If I understood correctly - instead of setting fixed price, you want to set a random price per shop item and remove the config option for this?

thelindat:
```lua
local function randomPrice(price, min, max)
    return math.ceil(price * (math.random(min, max)/100))
end
```

Something like this, then have to explicitly make it fluctuate.

frazz1234:
Could be very amazing .. 

But if an item is more expensive like in a shop then  other one, it can make some trouble. Like a holder more cheaper in the black market than réguler shop 🤷‍♂️

thelindat:
Will look into this at a future date - likely going to add more stuff for shops alongside the items API.

Issue #375, opened 2021-12-13
Weapons in vehicles and reloading issues


frazz1234:
So if you DriveBy and need to reload, r put yourself in cinematic mode. So something in this script for desactivate this could be amazing! thanks

dunak-debug:
Could you please test this native to disable cinematic mode?

https://docs.fivem.net/natives/?_0xDCF0754AC3D6FD4E

frazz1234:
SO the native work, but for anyreason... the gun is holstered and have to take it back

thelindat:
I'll look at drivebys another time, but it's always been awkward to get working properly.

Issue #534, opened 2022-02-09
Vehicles' weapons


0xDEMXN:
**Describe the bug**
Vehicles' weapons are not working at all, @DokaDoka hinted it *may be* something related to `SetWeaponsNoAutoswap`
I'll let this up to anyone who want to look into it.

PR #542, opened 2022-02-12
New ammo logic

0xDEMXN:
I don't know if you actually want this, this is not a breaking change but logic behind ammo changes a little bit.

- Ammos are now 'ammo packs' or 'magazines' that load up weapons based on a multiplier
- You can stack ammo in the weapon up to an hardcap
- Uses GTA standard reloading

Basically, no more single bullets being reloaded everytime you go from full clip to 0 and all the logic that came with it.
I did this because I personally like it a lot more but this also fixes:
#375 
#534


thelindat:
I specifically made the ammo system because I don't like loading a million bullets into a single gun.

thelindat:
I don't really like how it's being handled here. I'm working on this stuff now since it's become enough of a nuisance, but it's more than a few native changes (I need to re-enable the weapon wheel and disable inventory weapons when using a weaponised vehicle).

0xDEMXN:
> I don't really like how it's being handled here. I'm working on this stuff now since it's become enough of a nuisance, but it's more than a few native changes (I need to re-enable the weapon wheel and disable inventory weapons when using a weaponised vehicle).

Ok cool I'll let you the hard work then

PR #552, opened 2022-02-15
feat: disable action mode and firing without aiming

0xDEMXN:
"features nobody asked for and where to find them" in bookstores anytime soon™


client.lua
```diff
@@ -971,6 +971,9 @@ RegisterNetEvent('ox_inventory:setPlayerInventory', function(currentDrops, inven
 			if currentWeapon then
 				DisableControlAction(0, 140, true)
 
+				if not IsPlayerFreeAiming(PlayerData.id) then DisablePlayerFiring(PlayerData.id, true) end
```
thelindat, line 974: Maybe tie this to a new convar?

thelindat (changes requested):

PR #557, opened 2022-02-18
Add useful command for police (weapondetails)

lush0n3:


LukeWasTakenn:
I feel like this should be part of police job and not really the inventory

PR #562, opened 2022-02-19
Support limit when converting items from old db

root-dm:


thelindat:
This had some issues so I didn't add it, but I've removed limit anyway.

PR #564, opened 2022-02-20
fix(stash): export issue

PsychoShock:


LukeWasTakenn:
How does this fix the exports issue? Just start the inventory after loading the config, I don't see how this accomplishes anything.

PsychoShock:
If someone had the inventory before the new update, `config.cfg` was starting the inventory script. Now with the new update, you need to start the inventory in the `server.cfg`. Not a "issue", just for people that was using a custom `config.cfg`.

PR #578, opened 2022-03-01
improvement: necessary dependencies

hyprell:
Added two more dependencies to fxmanifest.lua, inventory doesn't work properly without these resources. Also it will force resources to be started before inventory to prevent additional errors and bugs.

### DIFF
--- a/fxmanifest.lua
+++ b/fxmanifest.lua
@@ -16,7 +16,6 @@ dependencies {
 	'/server:5181',
 	'/onesync',
 	'oxmysql',
-	'es_extended',
 	'ox_lib'
 }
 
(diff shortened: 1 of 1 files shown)


LukeWasTakenn:
ah yes, push commits going for standalone over the past month just to add es_extended as a dependency in the fxmanifest, nice.

hyprell:
when it will be standalone remove it, but as far as I know, it doesn't work standalone right now

thelindat:
> when it will be standalone remove it, but as far as I know, it doesn't work standalone right now... I will remove it, but ox_lib still needs to be included in fxmanifest.lua

I’ve been using the inventory without ESX for a month. Lib was removed from the manifest because a. I have my own dependency checks and b. Having to restart 4 resources on my server whenever I restart lib during development is a shitshow.

PR #580, opened 2022-03-01
Check item limit when swapping from player inventory

dolutattoo:
Just realized that there is a way to glitch the item limit.

There is actually 2 ways to swap 2 items when interacting with another inventory:
- If you swap an item with anothrer one, sarting from **another** inventory, the limit is correctly checked and everything is fine.
- If you swap an item with another one, starting from **your** inventory, you'll be able to carry more than the limit.

Here is the reproduce steps:
- Give yourself 1 `water` & 6 `testburger` (testburger **limit is 3**, but using giveitem allows you to carry more),
- Take a vehicle, and put 3 burgers in the trunk,
- Re-open the trunk, just to make sure,
- Try to swap the 3 `testburger` from the trunk, with the `water` = everything is fine, item check works,
- Try to swap your `water` with the 3 `testburger` of the trunk = you now have 6 `testburger`

So Im not sure if this is the correct way to fix it, but it worked fine for me.

### DIFF
--- a/server.lua
+++ b/server.lua
@@ -205,24 +205,39 @@ ServerCallback.Register('swapItems', function(source, data)
 				fromInventory = (data.fromType == 'player' and playerInventory) or Inventory(playerInventory.open)
 			end
 
-			if not sameInventory and toInventory.type == 'player' or toInventory.type == 'otherplayer' then
-				local fromData = fromInventory.items[data.fromSlot]
+			if not sameInventory then
+				if fromInventory.type == 'player' then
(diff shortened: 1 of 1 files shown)


thelindat:
Seems like there's a bit of unnecessary code here, so I'll go fix it up myself.

thelindat:
This whole limit system is a clusterfuck I never wanted to deal with.
This PR doesn't work when moving water from inventory to another. There's also a lot of excess code being triggered regardless of if the item has a limit.

PR #581, opened 2022-03-01
Create si.lua

PandaSLO:


data/locales/si.lua
```diff
@@ -0,0 +1,86 @@
+return {
+-- Buttons, context menu
+['ui_use'] = "Uporaba",
+['ui_give'] = "Daj",
+['ui_close'] = "Zapri",
+['ui_drop'] = "Spusti",
+['ui_removeattachments'] = "Odstrani priloge",
+-- Tooltip
+['ui_durability'] = "Trajnost",
+['ui_ammo'] = "strelivo",
```
LukeWasTakenn, line 10: Should this be lowercase?

data/locales/si.lua
```diff
@@ -0,0 +1,86 @@
...
+['weapon_hand_wrong'] = "Napačno orožje v roki",
+['open_player_inventory'] = "Odpri inventar igralca~",
+['open_secondary_inventory'] = "Odpri sekundarni inventar~",
+['disable_hotbar'] = "Prikaži vročo vrstico inventarja~",
+['reload_weapon'] = "Znova naloži orožje~",
+['use_hotbar'] = "Uporabi element vroče vrstice %s~",
+['no_durability'] = "Trajnost artikla je izčrpana",
+['cannot_give'] = "Cilju ni mogoče dati %s %s",
+['evidence_cannot_take'] = "Ni dovolj visoka ocena, da bi jemljejo iz dokazov",
+['dumpster'] = "smetnjak"
```
LukeWasTakenn, line 85: Should this be lowercase as well?

data/locales/si.lua
PandaSLO: yes

PandaSLO: in our language we would type that in lowercase

data/locales/si.lua
PandaSLO: yes

data/locales/si.lua
LukeWasTakenn: I have no idea how that makes any sense, but cool!

data/locales/si.lua
SloRunner: evidence_cannot_take is not right btw

data/locales/si.lua
SloRunner: ui_removeattachments is not right also

PR #596, opened 2022-03-06
feat(client): Custom context buttons

antond15:
Makes it possible to have multiple built-in interactions with items (_e.g. use rolling paper to make cigarette or joint_).
Preview: https://streamable.com/rytnia

### DIFF
--- a/web/src/dnd/onCustomButton.ts
+++ b/web/src/dnd/onCustomButton.ts
@@ -1,0 +1,5 @@
+import { fetchNui } from '../utils/fetchNui';
+
+export const onCustomButton = (id: number, slot: number) => {
+  fetchNui('useButton', { id, slot });
+};
--- a/web/src/dnd/onCustomButton.ts
+++ b/web/src/dnd/onCustomButton.ts
@@ -1,5 +1,0 @@
-import { fetchNui } from '../utils/fetchNui';
-
-export const onCustomButton = (id: number, slot: number) => {
-  fetchNui('useButton', { id, slot });
-};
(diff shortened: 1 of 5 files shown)


web/src/dnd/onCustomButton.ts
```diff
@@ -0,0 +1,5 @@
+import { fetchNui } from '../utils/fetchNui';
+
+export const onCustomButton = (id: number, slot: number) => {
+  fetchNui('useButton', { id, slot });
+};
```
LukeWasTakenn, line 5: Probably don't really need a separate file for this since it really wont be used outside the context menu.

web/src/typings/item.ts
```diff
@@ -4,4 +4,5 @@ export type ItemData = {
   usable: boolean;
   close: boolean;
   description?: string;
+  buttons?: any;
```
LukeWasTakenn, line 7: 😡

LukeWasTakenn (commented):
Looks good though, I like it.

web/src/typings/item.ts
antond15: I'm not sure how to do it differently.
I tried this, but it shows `Property 'length' does not exist on type '{ [key: number]: string; }'.` (_same with map_).
```ts
buttons?: {
  [key: number]: string;
};
```

Can't find any other solution 😕

LukeWasTakenn: I think it's because map expects an array, not an object with kvp. I could be wrong though.

antond15: Okay, that was easy 🤦🏾‍♂️ Learned something new.

PR #599, opened 2022-03-08
Restrict using hotbar items while dead

tony-stark-17:


dolutattoo:
Doesn't work for me, even if player is alive cause PlayerData is nil here.
image

thelindat:
Didn't realise the check was missing, but should be on useSlot and useItem functions rather than the hotkey.

PR #600, opened 2022-03-09
Show "Copy serial number" button only if serial exists

dolutattoo:
Without this condition, you can see the button on any weapons (e.g. WEAPON_SWITCHBLADE, which doesn't have any serial number, so it wont copy anything)

### DIFF
--- a/data/locales/fr.lua
+++ b/data/locales/fr.lua
@@ -5,7 +5,7 @@ return {
 	['ui_close'] = "Fermer",
 	['ui_drop'] = "Jeter",
 	['ui_removeattachments'] = "Retirer des accessoires",
-	['ui_copy'] = "Copy serial number",
+	['ui_copy'] = "Copier le numéro de série",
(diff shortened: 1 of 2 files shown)


LukeWasTakenn:
Wouldn't this cause issues where if a weapon doesn't have a serial number you can't remove components off of it?

Though I guess only weapons without serial number would be melee and melee weapons don't have components? 🤔

dolutattoo:
In fact, melee weapons doesn't have any component

LukeWasTakenn:
Cool, thank you Dolu ❤️

dolutattoo:
A better way to do it would be to let people do Ctrl+C and if there is no serial number, show an error notification

LukeWasTakenn:
> A better way to do it would be to let people do Ctrl+C and if there is no serial number, show an error notification

Yeh or avoid adding a new locale entry and just don't display anything kek

PR #601, opened 2022-03-10
Add a one Key option

zf-development:
Add the option of using only one key (F2).

Add this to the documentation:

setr inventory:oneKey false
sert inventory:keys ['F2', 'K', 'Z']
# If you only want to have one to control all inventories.
setr inventory:oneKey true

The inventory:keys will still need the 3 keys if error is made in the configs by the users.

### DIFF
--- a/client.lua
+++ b/client.lua
@@ -446,75 +446,32 @@ function OnPlayerData(key, val)
 end
 
 local function RegisterCommands()
+	if shared.oneKey then
+		RegisterCommand('inv', function()
+			if not invOpen then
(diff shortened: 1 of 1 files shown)


thelindat:
With few logic changes you can get pretty much the same result by just binding to the same key; I really don't like a single-key implementation though.

zf-development:
I tried to just put the same key and it actually doesn't work. Here's the reason why BUT I just got an idea and will change the code today. For something "easier" (no configuration required)

thelindat:
> I tried to just put the same key and it actually doesn't work

That's why I said it would require some changes.

zf-development:
This is currently being completly remake.

PR #613, opened 2022-03-16
containers can now reduce the weight of items inside

RedAlex:


LukeWasTakenn:
How would adding stuff to your container make you lighter? Doesn't make any sense.

RedAlex:
> How would adding stuff to your container make you lighter? Doesn't make any sense.

Carry a bag"container" with items inside(in real life) are more easy to handle.

dolutattoo:
The only advantage of a bag is the fact that you can easily carry something, not carry something heavier.
This is what the current containers item offers: more slots.
In my opinion, a lot of people are matrixed by these previous inventories where people can have 5+kg capacity while wearing a bag. But that was just because it was not slot based inventories. I think it's just perfect as it is right now

PR #617, opened 2022-03-17
Add support to custom weapon item name and add weapon tint functionality

pstomi:
1. Added support to able to use custom weapon item name like instead of WEAPON_SMG we can use item name smg. Many servers use this way and the current hardcoded WEAPON_ solution prevent them to use ox_inventory, because it would be really painful for them to migrate their existing database with custom weapon names to use new ox_inventory. 

2. Added functionality for the weapon tints, so servers wont need separate script for that. 

3. Bugfix for item effects

### DIFF
--- a/web/src/components/inventory/InventoryContext.tsx
+++ b/web/src/components/inventory/InventoryContext.tsx
@@ -65,7 +65,7 @@ const InventoryContext: React.FC<{
           <Item onClick={handleClick} data={{ action: 'drop' }}>
             {Locale.ui_drop}
           </Item>
-          {props.item.name.startsWith('WEAPON_') && props.item.metadata?.serial && (
+          {props.item.metadata?.serial && (
             <>
               <Separator />
               <Item
--- a/web/src/components/inventory/InventoryContext.tsx
+++ b/web/src/components/inventory/InventoryContext.tsx
@@ -65,7 +65,7 @@ const InventoryContext: React.FC<{
           <Item onClick={handleClick} data={{ action: 'drop' }}>
             {Locale.ui_drop}
           </Item>
-          {props.item.metadata?.serial && (
+          {props.item.hashname.startsWith('WEAPON_') && props.item.metadata?.serial && (
             <>
               <Separator />
               <Item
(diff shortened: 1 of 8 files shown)


web/src/components/inventory/InventoryContext.tsx
```diff
@@ -65,7 +65,7 @@ const InventoryContext: React.FC<{
           <Item onClick={handleClick} data={{ action: 'drop' }}>
             {Locale.ui_drop}
           </Item>
-          {props.item.name.startsWith('WEAPON_') && props.item.metadata?.serial && (
+          {props.item.metadata?.serial && (
```
LukeWasTakenn, line 68: How'd this work? Not all weapons have a serial 🤔

thelindat:
> Many servers use this way and the current hardcoded WEAPON_ solution prevent them to use ox_inventory, because it would be really painful for them to migrate their existing database

I don't really care.

web/src/components/inventory/InventoryContext.tsx
pstomi: Fixed, I added new change where it checks with hashname. @LukeWasTakenn

pstomi:
> > Many servers use this way and the current hardcoded WEAPON_ solution prevent them to use ox_inventory, because it would be really painful for them to migrate their existing database
> 
> I don't really care.

@thelindat   I know it is not your problem, but here is the solution what would help many people. Also in general it is never a good thing to use hardcoded stuffs in an actual code, it is not a good coding practice, it can lead to many problem and a code is harder to maintain.

thelindat:
> I know it is not your problem

Literally everything people try to push is my problem.

pstomi:
> > I know it is not your problem
> 
> Literally everything people try to push is my problem.

@thelindat  so no chance that this will be approved and merged?

thelindat:
I need to fix various issues I have with this or at least review piece by piece for you to fix.

thelindat:
I've implemented the functionality you wanted, so you should just need to PR your tints.

pstomi:
@thelindat   Thank you so much. I will open separate PR about the Tints.

PR #618, opened 2022-03-17
fix progress exemple

N-fire:


LukeWasTakenn:
These docs are already outdated and since they are being rewritten there's no need to fix it here.

PR #620, opened 2022-03-18
Fix inventory don't show item if slots are uper then shared.playerslots

RedAlex:
After convert from an old inventory, players can have item in a slot not valid.
Server global max slots are 20 but player have 30 slots after use convert.lua.
In normal condition the inventory Only show the first 20 slots But the weight of other item are count.

Here the result after this fix
FiveM_b2189_GTAProcess_UWvPcp4N7J

I have add too a reorder item if the player connect Only in the case of he have more slots then shared.playerslots
FiveM_b2189_GTAProcess_3avyBd6jm7
FiveM_b2189_GTAProcess_HhDUuZsotY

After have play and disconnect with an empty slot the code slowly reduice the inventory to shared.playerslots

### DIFF
--- a/client.lua
+++ b/client.lua
@@ -876,6 +876,13 @@ RegisterNetEvent('ox_inventory:setPlayerInventory', function(currentDrops, inven
 
 	while not uiLoaded do Wait(50) end
 
+	local maxSlots = shared.playerslots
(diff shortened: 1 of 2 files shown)


thelindat:
Ideally the player slots would decrease as items are removed, rather than relogging.

PR #624, opened 2022-03-24
One key support.

zf-development:
If both first and second key are the same, the checking is different for better support that it is currently.

Glovebox, Trunk, Markers, Drops and Regular inventory is working without troubles.

No Configuration, no documentation edit.


thelindat:
It really isn't necessary, nor good practice, to just duplicate the code like that.

PR #629, opened 2022-03-27
Add export to check if inventory is open

ffpetrozzi:


LukeWasTakenn:
`LocalPlayer.state.invOpen`

PR #630, opened 2022-03-27
Correction of Italian translation

ffpetrozzi:


LukeWasTakenn:
Dude we don't want the export for opening the inventory when you can just use statebags.

I don't understand why you are trying to smuggle it into the locale PR.

LukeWasTakenn:
I see you removed it in one the "Update client.lua" commits, either squash them or create a new PR just on the `it.lua` file.

ffpetrozzi:
Okay, i'm doing it rn

Il giorno dom 27 mar 2022 alle ore 11:44 Luke ***@***.***> ha
scritto:

PR #655, opened 2022-04-11
Added dependency ox_lib

Tinky124:


thelindat:
Dependencies are a pain in the arse thanks to dependent resources stopping when that resource does.
People running into the issue of ox_lib not being started typically run into the same issue of es_extended not being started (because they didn't read the documentation).

PR #695, opened 2022-05-02
fix for setjob errors (Markers)

lush0n3:
There's a little bug when switching jobs, and that's caused by the lack of this code.

image

When printing the blips, the last shop is returned as null.

image
image

After adding the code, the problem is gone. I hope it will be of great help, greetings.


thelindat:
Not a fix, but at least a clue to the cause.

Better than the usual no information or false information.

PR #709, opened 2022-05-09
fix (issue overextended#707 ): Check if price is lower than 0

PlouffeLuL:


LukeWasTakenn (approved):
Seems to solve the issue, perhaps should still limit the maximum input number in the UI?

thelindat:
I tried one zero up to 40 zeroes and never repro'd the bug 🤷

Limiting would still be good; though a sensible number for items would be a nuisance when dealing with items without weight (i.e. money).

PR #715, opened 2022-05-15
fix custom weight for glovebox

StopherTv:
This block of lines make glovebox not work when puting custom weight cause value still be trunk while using glovebox. 
I try it by changing this block lines in two part (1 for trunk and 1 for glovebox) and all work perfectly


StopherTv:
Pull request #716 linked with this one. You need make this 2 edit for fix this bug . Thank You

thelindat:
What?

PR #716, opened 2022-05-15
Fix custom maxWeight for glovebox

StopherTv:
This line make glovebox not work when puting custom weight. 
I try it by deleting this line and all work perfectly


StopherTv:
Pull request #715 linked with this one. You need make this 2 edit for fix this bug . Thank You

thelindat:
Removing code isn't a fix.

PR #720, opened 2022-05-20
fix (inventory/server): item swap in evidence

PlouffeLuL:
So i dont think this is the best place to add a check and i feel like it should be done higher but it also needs to be when a swap is detected otherwise it will block any inventory manipulation.
Might not be the best way to do it but still fixes the 'bug' for now.


thelindat:
I thought it was blocked in the client; so if anybody did manage to sent the event it would be a trigger?

PlouffeLuL:
Someone posted it in discord today so i tried and it worked. Only for the itemswap tho but yes it works without a trigger or any external. Just item swap evidences and it will swap properly.

thelindat:
https://github.com/overextended/ox_inventory/commit/368344f845002a50e8c1e686e23c82d0da3b78cd

PR #734, opened 2022-05-30
Fixed 'Search' functions only returning single items

Kanzia-WOT:
- searches were only returning one slot for any item searched, no matter how many individual items were in the inventory. 
- Issue was that the key value it was trying to set in the 'returnData[item]' table was always 1. 
- It has been changed to set the slot number as the key and return the data of all instances of the items.
- Changed the return values so all 'slots' searches return in the same format, for compatability reasons

### DIFF
--- a/modules/inventory/client.lua
+++ b/modules/inventory/client.lua
@@ -62,7 +62,7 @@ function Inventory.Search(search, item, metadata)
 				end
 			end
 		end
-		if next(returnData) then return items == 1 and returnData[item[1]] or returnData end
+		if next(returnData) then return returnData end
 	end
 	return false
 end
--- a/modules/inventory/client.lua
+++ b/modules/inventory/client.lua
@@ -54,7 +54,7 @@ function Inventory.Search(search, item, metadata)
 				if v.name == item then
 					if not v.metadata then v.metadata = {} end
 					if not metadata or table.contains(v.metadata, metadata) then
-						if search == 1 then returnData[item][#returnData[item]+1] = PlayerData.inventory[v.slot]
+						if search == 1 then returnData[item][v.slot] = PlayerData.inventory[v.slot]
 						elseif search == 2 then
 							returnData[item] += v.count
 						end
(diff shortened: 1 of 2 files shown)


thelindat:
> searches were only returning one slot for any item searched, no matter how many individual items were in the inventory.

```lua
local ox_inventory = exports.ox_inventory

local water = ox_inventory:Search('slots', 'water')
local count = 0

for _, v in pairs(water) do
	print(v.slot..' contains '..v.count..' water '..json.encode(v))
	count = count + v.count
end

print('You have '..count..' water')
print('~~~~~~')

count = ox_inventory:Search('count', 'water')
print('You have '..count.. ' water')
print('~~~~~~')

local items = ox_inventory:Search('slots', {'water', 'burger'})

if items then
	for name, data in pairs(items) do
		count = 0

		for _, v in pairs(data) do
			if v.slot then
				print(v.slot..' contains '..v.count..' '..name..' '..json.encode(v))
				count = count + v.count
			end
		end

		print('You have '..count..' '..name)
	end
end

print('~~~~~~')

items = ox_inventory:Search('count', {'water', 'burger'})

if items then
	for name, count in pairs(items) do
		print('You have '..count..' '..name)
	end
end
```

image

So changing the return data to be indexed by slot id instead of as an array is a breaking change for absolutely no reason. No clue how you got the idea that it _only returns one slot for an item_.

> key value it was trying to set in the 'returnData[item]' table was always 1

No, it sets the table size + 1. The table starts at size 0, and increases by 1 for each found slot. That's how #table + 1 works.

> It has been changed to ... return the data of all instances of the items

Which it already does.

> Changed the return values so all 'slots' searches return in the same format, for compatability reasons

While that _does_ make it easier when searching for an unknown number of items, this isn't a "compatibility" fix - it's a breaking change.

thelindat:
> Makes it easier if you have code that needs to read single, and multi item searches

```lua
local items = {
	{'water', 'burger', 'cola'},
	{'water'}
}

for i = 1, 2 do
	local count = exports.ox_inventory:Search('count', items[i])

	if type(count) == 'number' then
		print(count)
	else
		print(json.encode(count))
	end
end
```

Kanzia-WOT:
Ahh I see where I've gone wrong here, my bad. I apprecaite the response man, disregard the pr

PR #737, opened 2022-06-10
fix(client): use phone

N-fire:
disable npwd:setPhoneVisible when npwd is not started

### DIFF
--- a/modules/items/client.lua
+++ b/modules/items/client.lua
@@ -75,7 +75,9 @@ Item('parachute', function(data, slot)
 end)
 
 Item('phone', function(data, slot)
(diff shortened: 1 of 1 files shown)


LukeWasTakenn (approved):

thelindat:
I would prefer wrapping the exports with pcall so we can ignore the error, rather than checking resource state each time it executes.
Will do something later.

On a related note, it's weird that you can pcall an export directly when an error occurs while calling it, but not if the export doesn't exist. 🤔

Something like this _should_ work.
```lua
Item('phone', function(data, slot)
	local success, result = pcall(function() return exports.npwd:isPhoneVisible() end)

	if success then
		exports.npwd:setPhoneVisible(not result )
	end
end)
```

PR #739, opened 2022-06-12
feat(weapon): adding missing weapon's

GRZ-D3V:
Adding weapon's that are missing. Some are missing a image too.


GRZ-D3V:
Similar in style to the [Compact Grenade Launcher](https://www.gtabase.com/grand-theft-auto-v/weapons/compact-grenade-launcher), the Compact EMP Launcher fires unique EMP grenades that can disable the engine of a vehicle for 5 seconds (without damaging it).

PR #756, opened 2022-07-01
cn.lua

Tompizza:
Chinese translation simplified


LukeWasTakenn (changes requested):
Please properly create a `.lua` file for the translations and only include english characters in it's name.

PR #790, opened 2022-08-08
Remove Prints

Shorty:
Remove the prints from the new clothing item function.


thelindat:
I left them because wip

PR #791, opened 2022-08-10
fix(workflows): update actions

CoachYT1:


LukeWasTakenn:
Why though? Updating action is completely unnecessary if it already works well, which I'd say it does.

This is like asking to break it.

PR #804, opened 2022-08-20
Fix several issues with qb compatibility

BerkieBb:


modules/inventory/server.lua
```diff
@@ -816,7 +816,7 @@ function Inventory.RemoveItem(inv, item, count, metadata, slot)
 			local array = table.create(#slots, 0)
 
 			for k, v in pairs(slots) do
-				array[k] = {item = type(v) == 'number' and { slot = v } or v, inventory = inv.type}
+				array[k] = {item = type(v) == 'number' and { slot = v, label = item.label, name = item.name } or v, inventory = inv.type}
```
thelindat: This change should be removed and squashed.

thelindat (changes requested):

PR #829, opened 2022-09-03
tweak(modules/inventory/server.lua): modify AddItem

mahanmoulaei:
Now the correct amount of unique(not stack) items would be added to the inventory.

### DIFF
--- a/modules/inventory/server.lua
+++ b/modules/inventory/server.lua
@@ -639,8 +639,9 @@ function Inventory.AddItem(inv, item, count, metadata, slot, cb)
 	if item then
 		if inv then
 			local totalCount = count
-			metadata, count = Items.Metadata(inv.id, item, metadata or {}, count)
 			for i = 1, totalCount do
+				metadata, count = Items.Metadata(inv.id, item, metadata or {}, count)
+				
 				---@type boolean?
 				local existing = false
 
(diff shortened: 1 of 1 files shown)


thelindat:
Lazy to review and test rn; this is to add multiple items to different slots when an item is unique, rather than forcing a count of 1?
Items.Metadata would need to be called for each slot to ensure each item gets unique metadata; that function also forcibly sets a count of 1 in some cases.

mahanmoulaei:
>this is to add multiple items to different slots when an item is unique, rather than forcing a count of 1? 

Yes

----
> Items.Metadata would need to be called for each slot to ensure each item gets unique metadata

Ok I'll change the code

----
> that function also forcibly sets a count of 1 in some cases

Yes I'm aware of this and the way I looked at it, shouldn't be conflicting with my PR(hopefully)

thelindat:
I have a few issues with how this is done, and it's easier to update the functionality myself than talk through the changes I'd like to see.

mahanmoulaei:
> I have a few issues with how this is done, and it's easier to update the functionality myself than talk through the changes I'd like to see.

Sure Linden, do as you wish.

thelindat:
@mahanmoulaei https://github.com/overextended/ox_inventory/commit/7de222bdc27e978961c18faf8eb6b3f17e74ccdd

mahanmoulaei:
> @mahanmoulaei [7de222b](https://github.com/overextended/ox_inventory/commit/7de222bdc27e978961c18faf8eb6b3f17e74ccdd)

Perfect Thank you!!!

PR #859, opened 2022-09-19
items total weight

Frowmza:
total weight should be counted as total item count multiply item weight i guess?

### DIFF
--- a/web/src/helpers/index.ts
+++ b/web/src/helpers/index.ts
@@ -48,6 +48,6 @@ export const itemDurability = (metadata: any, curTime: number) => {
 };
 
 export const getTotalWeight = (items: Inventory['items']) =>
-  items.reduce((totalWeight, slot) => (isSlotWithItem(slot) ? totalWeight + slot.weight : totalWeight), 0);
+  items.reduce((totalWeight, slot) => (isSlotWithItem(slot) ? totalWeight + (slot.weight * slot.amount) : totalWeight), 0);
 
 export const isContainer = (inventory: Inventory) => inventory.type === InventoryType.CONTAINER;


thelindat:
Unless something changed, slot != item.
Slot is the data for the slot, which _contains_ item data, but the slot weight is `(item.weight + item.metadata.weight) * item.count`.

I'm also not certain where this gets used since the server reports the inventory weight to the client anyway. Did you actually _see_ an issue or decide this was problematic when going through code?

Frowmza:
then the totalWeight function should look like this 
```
const getTotalWeight = (items: Inventory['items']) =>
  items.reduce((totalWeight, slot) => (isSlotGotItem(slot) ? totalWeight + (slot.metadata?.weight ? (slot.weight + slot.metadata?.weight) * slot.amount : slot.weight * slot.amount) : totalWeight), 0);
```

thelindat:
> then the totalWeight function should look like this
> 
> ```
> const getTotalWeight = (items: Inventory['items']) =>
>   items.reduce((totalWeight, slot) => (isSlotGotItem(slot) ? totalWeight + (slot.metadata?.weight ? (slot.weight + slot.metadata?.weight) * slot.amount : slot.weight * slot.amount) : totalWeight), 0);
> ```

You're not listening. The slot weight is set on the server and sent to the client, which gets sent into the ui.

Frowmza:
and in stackitems you only get 1 item weight cause they have same weight and items with metadata wight cannot be stacked

```
targetInventory.items[toSlot.slot - 1] = {
        ...targetInventory.items[toSlot.slot - 1],
        amount: toSlot.amount + amount,
        weight: fromSlot.weight,
    };
 ```

PR #883, opened 2022-10-02
Feat: Support modular weapon attachments + change metadata component structure

PlouffeLuL:
This _shouldn't_ be breaking anything with the current weapons components as i used the 'checkMetadata' function to change it structure to the new one. Obviously the hash will be empty but removing then adding the component will fix it. It also shouldn't be an issue in any ways as the hash is not currently used internally the add/remove component function works pretty much the same.

Some modular addon weapons requires multiples components at the same time.
ex: Sniper scopes also needs a rail, some silencer also needs a barrel ect.
So being able to add multiple component with only using one is a must in those situations.

The metadata structure change is because when trying to access weapon components from another resource its really annoying to retrieve the component hash used.
ex: When trying to recreate the weapon as a weapon object.

### DIFF
--- a/client.lua
+++ b/client.lua
@@ -318,32 +318,45 @@ local function useSlot(slot)
 				local components = data.client.component
 				local componentType = data.type
 				local weaponComponents = PlayerData.inventory[currentWeapon.slot].metadata.components
-				-- Checks if the weapon already has the same component type attached
-				for componentIndex = 1, #weaponComponents do
-					if componentType == Items[weaponComponents[componentIndex]].type then
+				local newComponents = {}
+				-- -- Checks if the weapon already has the same component type attached
+				for k,v in pairs(weaponComponents) do
+					if componentType == Items[v.name].type then
 						-- todo: Update locale?
 						return lib.notify({ type = 'error', description = shared.locale('component_has', data.label) })
 					end
 				end
+
+				-- This is for modulars addon weapons, lots of them requires multiples components at the same time
+				-- ex: Sniper scopes also needs a rail, some silencer also needs a barrel ect.
 				for i = 1, #components do
 					local component = components[i]
-
 					if DoesWeaponTakeWeaponComponent(currentWeapon.hash, component) then
 						if HasPedGotWeaponComponent(playerPed, currentWeapon.hash, component) then
-							lib.notify({ type = 'error', description = shared.locale('component_has', data.label) })
-						else
-							useItem(data, function(data)
-								if data then
-									GiveWeaponComponentToPed(playerPed, currentWeapon.hash, component)
-									table.insert(PlayerData.inventory[currentWeapon.slot].metadata.components, data.name)
-									TriggerServerEvent('ox_inventory:updateWeapon', 'component', tostring(data.slot), currentWeapon.slot)
-								end
-							end)
+							return lib.notify({ type = 'error', description = shared.locale('component_has', data.label) })
 						end
-						return
+						table.insert(newComponents, component)
 					end
 				end
-				lib.notify({ type = 'error', description = shared.locale('component_invalid', data.label) })
+
+				if #newComponents < 1 then
+					return lib.notify({ type = 'error', description = shared.locale('component_invalid', data.label) })
+				end
+
+				useItem(data, function(data)
+					if data then
+						local component_hashs = {}
+						for i = 1, #newComponents do
+							local component = newComponents[i]
+							GiveWeaponComponentToPed(playerPed, currentWeapon.hash, component)
+							table.insert(component_hashs, component)
+						end
+
+						table.insert(PlayerData.inventory[currentWeapon.slot].metadata.components, {name = data.name, hash = component_hashs})
+
+						TriggerServerEvent('ox_inventory:updateWeapon', 'component', tostring(data.slot), currentWeapon.slot, PlayerData.inventory[currentWeapon.slot].metadata.components)
+					end
+				end)
 			elseif data.allowArmed then
 				useItem(data)
 			end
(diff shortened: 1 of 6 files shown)


PlouffeLuL:
Also my commits are shitty and you can ignore the breaking change tag i fixed it, it wouldn't make any sense to completely break components for this.

modules/inventory/server.lua
```diff
@@ -1533,13 +1533,13 @@ RegisterServerEvent('ox_inventory:updateWeapon', function(action, value, slot)
...
 				if type == 'number' then
-					Inventory.AddItem(inventory, weapon.metadata.components[value], 1)
+					Inventory.AddItem(inventory, weapon.metadata.components[value].name, 1)
 					table.remove(weapon.metadata.components, value)
 					weapon.weight = Inventory.SlotWeight(item, weapon)
 				elseif type == 'string' then
 					local component = inventory.items[tonumber(value)]
 					Inventory.RemoveItem(inventory, component.name, 1)
-					table.insert(weapon.metadata.components, component.name)
+					weapon.metadata.components = data
```
thelindat, line 1542: Don't just take data from the client and update components with that; should be inserting like it was before.

thelindat (changes requested):

modules/inventory/server.lua
PlouffeLuL: Not sure how this would be done other then getting the data from client, i could add a second check server side to validate all the data passed but other then that i would mean changing the logic of adding components from client to server.

Would data validation be enough ?

PlouffeLuL: I added a server side validation that is basically recreating the component table.
I initially wanted to used inventory.weapon and call ``DoesWeaponTakeWeaponComponent`` / ``HasPedGotWeaponComponent`` but those arent rpc.

client.lua
```diff
@@ -318,32 +318,45 @@ local function useSlot(slot)
...
-								if data then
-									GiveWeaponComponentToPed(playerPed, currentWeapon.hash, component)
-									table.insert(PlayerData.inventory[currentWeapon.slot].metadata.components, data.name)
-									TriggerServerEvent('ox_inventory:updateWeapon', 'component', tostring(data.slot), currentWeapon.slot)
-								end
-							end)
+							return lib.notify({ type = 'error', description = shared.locale('component_has', data.label) })
 						end
-						return
+						table.insert(newComponents, component)
```
Waves966, line 338: wtf plouffe you stupid frenchman

client.lua
```diff
@@ -318,32 +318,45 @@ local function useSlot(slot)
...
+					return lib.notify({ type = 'error', description = shared.locale('component_invalid', data.label) })
+				end
+
+				useItem(data, function(data)
+					if data then
+						local component_hashs = {}
+						for i = 1, #newComponents do
+							local component = newComponents[i]
+							GiveWeaponComponentToPed(playerPed, currentWeapon.hash, component)
+							table.insert(component_hashs, component)
```
Waves966, line 352: again?

client.lua
```diff
@@ -318,32 +318,45 @@ local function useSlot(slot)
...
+						end
+
+						table.insert(PlayerData.inventory[currentWeapon.slot].metadata.components, {name = data.name, hash = component_hashs})
```
Waves966, line 355: a third time?

client.lua
```diff
@@ -1172,7 +1185,7 @@ RegisterNUICallback('removeComponent', function(data, cb)
 			if HasPedGotWeaponComponent(cache.ped, currentWeapon.hash, component) then
 				RemoveWeaponComponentFromPed(cache.ped, currentWeapon.hash, component)
 				for k, v in pairs(itemSlot.metadata.components) do
-					if v == data.component then
+					if v.name == data.component then
 						table.remove(itemSlot.metadata.components, k)
```
Waves966, line 1189: why table remove

Waves966 (changes requested):
stop being shit

LukeWasTakenn:
why was this closed?

thelindat:
@PlouffeLuL Don't know why you closed it but finally trying to review and test this. Need to know exactly what this is fixing.

Testing `WEAPON_HEAVYSNIPER_MK2` (with your fork) and can only attach one of suppressor, barrel, or muzzle; I assume that's what's supposed to change?

PlouffeLuL:
Not really a 'fix' as there is nothing to be fixed initially, more of a tweak. 

```
		['at_scope_advanced'] = {
			label = 'Advanced Scope',
			type = 'sight',
			weight = 280,
			client = {
				component = {`COMPONENT_MCXSPEAR_SCOPE_18`,`COMPONENT_MCXSPEAR_RAIL_2` },
				usetime = 2500
			}
		},
```

Currently if you use this only the 'scope' component will be added to the weapon, for regular weapons its perfectly fine and work as it should be.
When you're using addon weapons some of them are considered 'modular' basicly just means they have a fuck ton of attachements. 
Some of them like the mcx spear used in the example are missing some parts when adding a scope, in the example its the scope rail.
So if the current version of the inventory if i add a scope on the mcx its floating over the gun, with the tweak a added it will also add the rail.

For the other part the idea was to allow the ability for people working with the inventory to know exactly what component hash is attached to the weapon. 
Currently if i get the weapons metadata the attachements on the weapon will return the item name and not the attachement name/model. 
I wanted to change that when i made my weapon script so i can also show the weapon components on the weapon when its on your back. If i you're using a weapon object and easy work around would be setting the component on the weapon the exact same way its done in the inventory but as i decided the use weapon entities instead of objects for different reasons it started being annoying cause the natives like 'DoesWeaponTakeComponent' and such obviously wont work on not weapon entities.
Also i figured if Qb-Inventory already as this features why not ox.

I think i found a small issue when login in and out tho, i **think** the component metadata structures gets fucked by the checkmetata ... check ? i added. It was added to make sure that when loading an inventory the component structure was being changed to the new one but i think it somehow also affect the current one. Not completely sure if its me breaking something or an actual issue, i would need to check into it.

thelindat:
Changing the component table is fairly breaking for 3rd party, and I don’t really like how the data is being assigned. Probably best to just assign the component hashes to a separate tables.

If we do go with breaking compatibility though I’d rather do a bigger change to improve weapon and component declarations. There’s plenty of weapons missing and countless components don’t exist or mismatch.

PlouffeLuL:
Yea, i also dont really like the way data is assigned. Tried to keep as much as the original code i could but the weapons / component is currently not in the best state to be work with. I agree with you if this was to be added might as well be in a much bigger update.

thelindat:
> I agree with you if this was to be added might as well be in a much bigger update.

Leave this PR here for reference. I have a million other things I need to work on but haven't been motivated to do so, so weapons improvements is just another.

PR #885, opened 2022-10-05
fix(weapons): heavyrifle ammo

izMystic:


izMystic:
doesnt seem like it does anything to fix having infinite ammo with it

PR #899, opened 2022-10-14
Finnish translation

EpKouhia:
Finnish translations for the inventory.


eprimex:
Word "inventory" cannot be used as part of Finnish language. Please change them to proper Finnish words

AdrineX:
> Word "inventory" cannot be used as part of Finnish language. Please change them to proper Finnish words. Ideal Finnish words are like varasto (as a place to store items in) , tavarajärjestelmä (inventory as a system)

Indeed, inventory does not translate into Finnish properly at anyway, but people are used to call it that way in most of the Finnish servers and it'd say it's the best way to "translate" it to Finnish and use in normal communication. "Varasto" also is not a proper way to use the translation as it translates to English as "Storage" and nobody carries a storage around with a player. More proper way would be. As i said there is no proper way to call inventory in Finnish language, more proper way would be reppu = backpack or taskut = pockets, but there could be item for a Backpack that has its own inventory so I would say the best way to describe inventory in Finnish in the translation as it is now as everyone knows in Finnish communities what it means. I don't think we need to be grammaly correct in this one.

PR #906, opened 2022-10-19
Tell me im wrong

Testaross:
This fixes my on client add issues, who the fuck knows


Testaross:
When you have an item like 
```
['largebag'] = {
            label = 'Large Duffle Bag',
            weight = 220,
            stack = false,
            consume = 0,
            ammoname = 'ammo-rifle',
            client = {
                add = function(total)
                    if total > 0 then
                        print('added but weird')    
                    end
                end,
                remove = function(total)
                    if total < 1 then
                    print('thefuck')
                    end
                end
            }
        },```
And you buy it from a store, it just fucks up the item until restart because it uses set slot, i swapped it to additem and now it correctly passes the function

PR #909, opened 2022-10-20
fix(client): Update Reload weapon keybind name

jag3dagster:
* ~~Appears that pefcl has a conflicting keybind which was~~ Players are unable to reload with default keybind 'R'. Updating the keybind name appears to resolve the issue.

### DIFF
--- a/client.lua
+++ b/client.lua
@@ -590,7 +590,7 @@ local function registerCommands()
 	})
 
 	lib.addKeybind({
-		name = 'reload',
+		name = 'reloadweapon',
 		description = locale('reload_weapon'),
 		defaultKey = 'r',
 		onPressed = function(self)


thelindat:
I'm confused, where is pefcl registering a 'reload' command? This is also the same command / key mapping that has been used for a year (just moved to the ox_lib api).

jag3dagster:
Hmm, I may have been hasty with the assumption it was a conflict with pefcl as all I'm seeing is a dictionary that has 'reload' listed as key.

Honestly, I have no idea why it broke. Potentially a caching issue? Not entirely sure, I just know that after updating to 2.13.2, I was unable to reload weapons when pressing 'R', though I could reload by using an ammo item. Everything looked right in keybind settings and I couldn't find any errors anywhere. Once I swapped the name to 'reloadweapon' I was able to reload using the 'R' keybind again.

jag3dagster:
Additionally, the command used to be `reload` now, with the ox_lib api, it ends up being `+reload` could have something to do with it as well.

thelindat:
image

Classic weird fayoum.

PR #914, opened 2022-10-21
feat(client): item skillcheck

SamShanks1:
not tested yet cuz im lazy
will add to docs if its merged
waiting for linden to change lua as he does for all my prs


thelindat:
Doesn't really add anything / make it easier to handle skill checks. I might not even do the progress bar today but it's there for good.

PR #923, opened 2022-10-25
better blips method [UNTESTED]

MrNewb:
theres apparently a string limit and the fivem discord has a conversation with bdub saying this is better. I dont have something to test it with currently but figured a pr with it will do. It works fine in qb with everything I throw at it.


MrNewb:
its still a string I guess but supposedly its better

PR #927, opened 2022-10-28
Fix access to glovebox and trunk

Cyril-Beeckman:
Since the new code to vehicle access, we can't open glovebox or trunk


thelindat:
Could have sworn I changed that 😞

PR #934, opened 2022-11-02
I provide language support for Simplified Chinese cn.lua

442516:
Today I am offering a simplified Chinese locale because I see most individual Chinese players using ox_inventory. Think it's cumbersome to edit the locale. So today I offer one.


thelindat:
This should be `locales/zh-cn.json`, you're using the old Lua format (and a translated directory name).

PR #945, opened 2022-11-08
Allow GetItem to return Metadata and item slot

Dhawgy:
I recently switched from QBCore inventory and found this export to be lacking, I might simply be wrong and there's a better way to do this but I've been using this for a few days and found no issues so thought I'd share.


PlouffeLuL:
https://overextended.github.io/docs/ox_inventory/Functions/Server/Inventory/Search

If you use search you can already get those info

thelindat:
GetItem doesn’t return an item in a single slot, it just returns generic item data and the total count.

thelindat:
Closing this. If there's an issue with returns when using the QB functions, change the qb bridge.

PR #956, opened 2022-11-10
feat(client/server/web):  Image URL metadata

renzuzu:
i think its a cool feature to support Image Urls.
so many use cases.

this will support metadata.imageurl along with metadata.image.

metadata.image is still the first priority when checking condition.

### DIFF
--- a/client.lua
+++ b/client.lua
@@ -779,8 +779,7 @@ RegisterNetEvent('ox_inventory:updateSlots', function(items, weights, count, rem
 		if not item.name then
 			item = PlayerData.inventory[item.slot]
 		end
-
-		Utils.ItemNotify({item.metadata?.label or item.label, item.metadata?.image or item.name, removed and 'ui_removed' or 'ui_added', count})
+		Utils.ItemNotify({item.metadata?.label or item.label, item.metadata?.image or item.metadata?.imageurl or item.name, removed and 'ui_removed' or 'ui_added', count})
 	end
 
 	updateInventory(items, weights)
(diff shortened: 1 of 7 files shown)


thelindat:
I don't really want `imageurl` as a metadata field.
Probably the most obvious issue that could arise is servers allowing players to make items with their own images.
Also just obnoxiously long url paths that start slowing down inventory saving due to serialisation because somebody has hundreds of them in different slots.

renzuzu:
>Probably the most obvious issue that could arise is servers allowing players to make items with their own images.

whats wrong this? is it a choice of server owners?
this is the main reason why i created this push.

>Probably the most obvious issue that could arise is servers allowing players to make items with their own images.

What if we support only imgur url for example which is short url (which is comparable to itemname length)
ex. https://i.imgur.com/89c9gof.png

or use only Image id like `89c9gof` or `89c9gof.png` for pattern check in when converting it to full url.
the length is very similar to itemnames.

logic will be
```
backgroundImage: `url(${`${item.metadata?.image ? `${imagepath}/${item.metadata.image}.png` : item.metadata?.imageurl ? `https://i.imgur.com/${item.metadata.imageurl}` : `${imagepath}/${item.name}.png`}`})`,
```

anyway have a gooday, its your call

thelindat:
> whats wrong this? is it a choice of server owners?

Because people _will_ have images of pornography, gore, and other questionable content; as well as other potential issues of arbitrary urls (if not using a whitelist). Still not something I want to support at all.

PR #959, opened 2022-11-11
inventory:itemnotify convar to disable NUI Item update notification

ESK0:


thelindat:
This sounds horrible; why?

ESK0:
> This sounds horrible; why?

I believe that many people like me prefer minimalism. It is also suitable for a lot of serious RP servers, where notifications of a similar type are non-wl shit

thelindat:
> suitable for a lot of serious RP servers

In real life, when somebody hands me something or I pick something up, I don't need to look through my pockets and try to work out what I just got..

I do want a more minimalist style in the future, but removing them entirely is bizarre; yes I get it's an optional convar but 😕.

ESK0:
It is entirely up to you, I believe that a lot of people like a similar style notification and that's why I left it on by default.

I also believe that everyone will notice, for example, that he is holding a gun or that when he reloads the gun, it is more than clear that the amount of bullets in the inventory has been reduced.

PR #983, opened 2022-11-20
Skillcheck enabled crafting

xyroot:
Modified ox_inventory:startCrafting in client.lua to accept skill check options in data/crafting.lua, which will continue crafting normally if no skillcheck difficulty is indicated, and will show the progressbar after the skillcheck so it can still be cancelled.

May not be the best implementation due to the repeated callout of the lib.progressCircle, feedback is appreciated.

### DIFF
--- a/client.lua
+++ b/client.lua
@@ -1350,7 +1350,7 @@ end)
 
 lib.callback.register('ox_inventory:startCrafting', function(id, recipe)
 	recipe = client.craftingBenches[id].items[recipe]
-	if not recipe.difficulty then
+	if not recipe.difficulty  then
 		return lib.progressCircle({
 			label = locale('crafting_item', Items[recipe.name].label),
 			duration = recipe.duration or 3000,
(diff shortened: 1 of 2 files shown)


thelindat:
Weird code duplication; this could be as simple as
```lua
if recipe.difficulty and not lib.skillCheck(recipe.difficulty) then return end
```

Would also be nicer having randomly triggering skill checks and the progress bar "pause" and continue or some sort of progress/skillcheck hybrid. @LukeWasTakenn

xyroot:
Made a small change patterned over your simpler approach, but I guess I'm missing on how to not duplicate the calling of the progressCircle

Randomly triggering skill checks or pausing the progress bar would be a nice to have as well
Also integrating animations to skill checks, if that is something that can be looked into in a future update

thelindat:
As I said this is a one liner, literally just
```lua
lib.callback.register('ox_inventory:startCrafting', function(id, recipe)
	recipe = client.craftingBenches[id].items[recipe]

	if recipe.difficulty and not lib.skillCheck(recipe.difficulty) then return end

	return lib.progressCircle({
		label = locale('crafting_item', Items[recipe.name].label),
		duration = recipe.duration or 3000,
		canCancel = true,
		disable = {
			move = true,
			combat = true,
		},
		anim = {
			dict = 'anim@amb@clubhouse@tutorial@bkr_tut_ig3@',
			clip = 'machinic_loop_mechandplayer',
		}
	})
end)
```

But I'd rather look at this is the future after some possible changes to allow a hybrid progressbar/skillcheck.

PR #988, opened 2022-11-22
feat (crafting): add RegisterWorkStation export

PlouffeLuL:


PlouffeLuL:
This is based on the registerShop export, downside i see is that it makes the callback bigger but if its fine for shops im assuming its fine for crafting

PlouffeLuL:
Im doing as much testing i can currently and fixing issues as i see them, it should be fine now. Let me know if you see anything i can change.

thelindat:
I don't want the export since 3rd party integration isn't crucial for crafting at this time and I don't want to add a temporary method like I did for shops.

PR #996, opened 2022-11-27
An attempt at adding qb status

BerkieBb:


thelindat:
Should be able to make this trigger directly on the server to be more secure; especially since right now you could set any random metadata property.

MenacingMethodz:
@BerkieBb I just added this to the modules/bridge/qb/client.lua 

`function client.setPlayerStatus(values)
	for name, value in pairs(values) do
		TriggerServerEvent('QBCore:Server:SetMetaData', name,
			QBCore.Functions.GetPlayerData().metadata[name] + math.random(value))
	end
end`

This worked for me, let me know if this is a good direction that can help you guys out.

BerkieBb:
QBCore:Server:SetMetaData is a deprecated event and exploitable, I won't use that

BerkieBb:
> Should be able to make this trigger directly on the server to be more secure; especially since right now you could set any random metadata property.

Where could I do that? I searched for anything including the word `status` but only found useItem on the client side triggering it.

BerkieBb:
> > QBCore:Server:SetMetaData is a deprecated event and exploitable, I won't use that
> 
> It doesn't state on the QBCore Documentations that this is true.

Seems like when I was in the dev team I forgot to add the comment to that event, because it is exploitable

thelindat:
> Where could I do that? I searched for anything including the word `status` but only found useItem on the client side triggering it.

Right now the only event triggered after successfuly item use is
image

I could probably clean up the function a bit and use a client event callback so the server is always aware if the item was truly used or not; adds some extra security and we can trigger the status change there.

Lack of foresight also means that default items status are based on ESX max status, which is usually 1000000 rather than just 100.

DjKillerMemeStar:
Sorry to open this again but this code does not update the hud.

You need trigger the 'hud:client:UpdateNeeds' event for that. In my opinion that should be in the QB setmetadata function. I can PR it in either QB or here.

PR #1003, opened 2022-12-01
fix(modules/inventory): return false if inventory isn't loaded

hyprell:


thelindat:
better yet don't search when it isn't loaded, and don't do - what I assume you're doing - a while loop on the search while waiting for it to return a truthy value.

PR #1005, opened 2022-12-01
feat(modules/inventory/server): CanCarryWeight

hyprell:
Useful thing for creating scripts with metadata weight (filling bottles, buckets and etc.) to not exceed max inventory weight.


thelindat:
Could you update docs for this? <https://github.com/overextended/overextended.github.io/tree/main/docs/ox_inventory/Functions/Server/Inventory>

PR #1006, opened 2022-12-01
formatting and weapon switching bug fix

hyprell:


thelindat:
Random styling included and 6 commits that I need to manually squash, bit annoying.
I've fixed this myself without adding the second Disarm call.

PR #1013, opened 2022-12-04
fix(client/ox): only reduce value if more than 100

dolutattoo:


thelindat:
If the status value is 1000 or -1000, it's setup for ESX's system which is out of 1000000.

dolutattoo:
Oh yes, forgot the negative numbers. Current condition is still broken tho

PR #1014, opened 2022-12-04
fix(client/ox): status value

SamShanks1:
Use the correct greater/less than sign for assigning status values


SamShanks1:
https://discord.com/channels/813030955598086174/951683419769831465/1049100327032471623

SamShanks1:
Nvm @thelindat im confused with status

N-fire:
Why my satus = confused

thelindat:
Welcome to working with ESX status, which has it's values flipped around.
1000000 hunger would mean you're _super hungry_, right? Nope - you're full.

If an item adds 10000 hunger then it needs to actually remove 1 hunger.

PR #1042, opened 2022-12-15
Fix issue with buying license, used label instead of the actual license names, and added license price to dialog

cjosephdc:
- Fix issue with buying licenses
- Used label instead of the actual license name
- Added license price to dialog


thelindat:
Either keep the scope limited or use proper conventional commits. As for adding label, there was intent to get that using the _proper_ framework way - GlobalState for ox_core, database for esx; qb ruined that because lol no labels.

PR #1049, opened 2022-12-18
refactor(bridge/nd): licenses & groups

Andyyy7666:


modules/bridge/nd/client.lua
```diff
@@ -1,7 +1,41 @@
-RegisterNetEvent("ND:setCharacter", function(character)
-    local groups = {}
-    for group, info in pairs(character.groups) do
-        groups[group] = info.lvl
+function reorderGroups(groups)
```
thelindat, line 1: `local function`

modules/bridge/nd/client.lua
```diff
@@ -1,7 +1,41 @@
...
+end)
+
+RegisterNetEvent("ND:jobChanged", function(job, lastJob)
+    local character = NDCore.Functions.GetSelectedCharacter()
+    local groups = reorderGroups(character.data.groups)
+    groups[lastJob.name] = nil
+    groups[job.name] = job.rank
+    
+    OnPlayerData("groups", groups)
+end)
```
thelindat, line 31: You have multi-groups but only update "job", and remove lastJob?

modules/bridge/nd/server.lua
```diff
@@ -8,6 +8,13 @@ end)
 
 AddEventHandler("ND:characterUnloaded", playerDropped)
 
+RegisterNetEvent("ND:jobChanged", function(source, job, lastJob)
```
thelindat, line 11: No "reorderGroups" for server?

modules/bridge/nd/server.lua
```diff
@@ -82,3 +99,26 @@ function server.setPlayerData(player)
...
+function server.buyLicense(inv, license)
+	if server.hasLicense(inv, license.name) then
+		return false, "has_weapon_license"
+	elseif Inventory.GetItem(inv, "money", false, true) < license.price then
+		return false, "poor_weapon_license"
+	end
+
+	Inventory.RemoveItem(inv, "money", license.price)
+	NDCore.Functions.CreatePlayerLicense(inv.owner, "weapon")
+	return true, "bought_weapon_license"
```
thelindat, line 123: Outdated locales?

thelindat (changes requested):

modules/bridge/nd/client.lua
Andyyy7666: when job is set/changed then it will also set a group as that job

PR #1052, opened 2022-12-22
feat(web): Metadata image URL support (png & apng)

uShifty:
This pull request introduces the ability for script creators to include images in their scripts by specifying a web-hosted URL. This feature can be useful in a variety of situations, such as a business system that allows users to create new menu items on-demand. As an example, I have added a new item to the default web development inventory that demonstrates how to use this feature, but I can remove it if desired. Please let me know if you have any questions or concerns about this change.

### DIFF
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -48,6 +48,10 @@ debugData([
             metadata: { description: 'Generic item description' },
           },
           { slot: 5, name: 'water', weight: 100, count: 1 },
+          { slot: 6, name: 'backwoods', weight: 100, count: 1, metadata: {
+            label: 'Russian Cream',
+            image: "https://i.imgur.com/2xHhTTz.png"
+          }},
         ],
       },
       rightInventory: {
--- a/web/src/components/inventory/InventoryHotbar.tsx
+++ b/web/src/components/inventory/InventoryHotbar.tsx
@@ -31,7 +31,7 @@ const InventoryHotbar: React.FC = () => {
           <div
             className="hotbar-item-slot"
             style={{
-              backgroundImage: `url(${checkimage(item)})`,
+              backgroundImage: `url(${checkimage(item.metadata?.image ? item.metadata.image : item.name)})`,
             }}
             key={`hotbar-${item.slot}`}
           >
--- a/web/src/components/inventory/InventoryHotbar.tsx
+++ b/web/src/components/inventory/InventoryHotbar.tsx
@@ -6,7 +6,7 @@ import WeightBar from '../utils/WeightBar';
 import { Slide } from '@mui/material';
 import { useAppSelector } from '../../store';
 import { selectLeftInventory } from '../../store/inventory';
-import { imagepath } from '../../store/imagepath';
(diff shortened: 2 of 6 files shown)


thelindat:
To quote the last time this was PR'd.

> Because people will have images of pornography, gore, and other questionable content; as well as other potential issues of arbitrary urls (if not using a whitelist). Still not something I want to support at all.

To even consider this I'd require
- a different metadata property (like imageurl, as last suggested)
- whitelisted urls
- logging when an item is created with an image (this would, finally, be a decent use-case for Discord)

Some sort of content moderation tool would be useful to detect NSFW/NSFL content would also be desirable, though somewhat out of scope I expect people to just blindly accept user-input.

uShifty:
> To quote the last time this was PR'd.
> 
> > Because people will have images of pornography, gore, and other questionable content; as well as other potential issues of arbitrary urls (if not using a whitelist). Still not something I want to support at all.
> 
> To even consider this I'd require
> 
> * a different metadata property (like imageurl, as last suggested)
> * whitelisted urls
> * logging when an item is created with an image (this would, finally, be a decent use-case for Discord)
> 
> Some sort of content moderation tool would be useful to detect NSFW/NSFL content would also be desirable, though somewhat out of scope I expect people to just blindly accept user-input.

Personally I think a new meta field is unnecessary for something as simple as this any particular reason a new meta property would be better?  As to me its just a mess copying the image implantation to add support for URLs

Whitelisted urls is not problem, ill probably just add imgur to the regex as that is one of the only image hosts that use short urls.

Discord logs should be something that the script developer should implement into their resource when creating items, I dont see why the inventory should be moderating within the inventory. If you are wanting this to be implemented I can do that. Although I really think it should be something done from the script developers. Especially since the QBCore create item doesnt log so they are logging it within the resources that hook create item already.

thelindat:
> logs should be something that the script developer should implement into their resource when creating items

They won't, and this is specifically for assisting in moderation of images with unsafe content. I don't want people creating images with Nazi imagery, pornography, gore and outright depictions of death - and they _will do that_ unless something is in place. I don't trust anybody to properly secure this shit at all.

> why the inventory should be moderating within the inventory

Because when people can do the wrong thing, they will. Don't pretend people won't just make an event that lets people set the imageurl of any item they're holding.

> to me its just a mess copying the image implantation to add support for URLs

How is it "copying" the implementation? Rather then testing every slot for a whitelisted url, you simply do check if a property exists.

uShifty:
I will reopen a pull request with the requested changes after some testing

uShifty:
> logging when an item is created with an image (this would, finally, be a decent use-case for Discord)

Do you want me to implement a function to utilize discord webhooks? You mentioned this would be a good use for discord but, it wouldnt match all the other logs. Should I convert all the other logs to discord? Or just use lib.logger?

thelindat:
> it wouldnt match all the other logs

Not supposed to.

> Should I convert all the other logs to discord

... No.

The point of this is to say "an item has a user-submitted image - here's a link/embed of said image"; so you can at least _try_ to moderate it.

PR #1057, opened 2022-12-23
Replace shared.resource with cache.resource

TheFarFantomas:


thelindat:
I left it as-is because I don't know if people are relying on it in personal changes and it makes almost no difference.
When (and if) I do 3.0.0 stuff like this will get changed.

Issue #422, opened 2022-01-01
Support Peds for Shops!


TlYPN:
**Is your feature request related to a problem? Please describe.**
My feature request isn't directly a problem! and people could just use there 5 head! but i have noticed it hard for new players to figure out how to "open" the shop. without being told to use Qtarget on the register.

**Describe the solution you'd like**
Possibly support the use of a ped in the shop, as oppposed to "qtarget to a cash register"

**Describe alternatives you've considered**
GIANT HOLOGRAPHIC text to tell players to third eye cash register (im only kidding, i dont really have another solution)

**Additional context**
Sorry if this isnt the place for suggestions! wasnt sure where else to put it

thelindat:
It's not something any of us plans to actively work on (original bt-target support was a PR), but it's not undesired.
Can refer to https://github.com/overextended/ox_inventory/pull/321

I could setup some ped management as an export inside pe-lualib so it can handle peds for every resource (I need to move marker code there, too). As for actual implementation into the inventory and setting up all the positions, I'd probably just do one as a sample and leave the rest for community commits.

Senlar:
I found another resource called droopies-peds and I modified it to work with qtarget. I use it to manage all interactable peds on my server from one resource like your suggesting Linden. Here's my version: https://github.com/Senlar/droopies-peds

Also, could I please get unbanned from the discord? Haven't been able to find a way to contact anyone for a while.

Issue #1034, opened 2022-12-14
Shop marker not working if target is true


Gregory-Hepicloud:
**Describe the bug**
Create shop wihout target not working if setr inventory:target true in config

**To Reproduce**
1. Go to shop.lua
2. Create shop without option target or model
3. restart
4. See any shop show on map and no marker
5. go to ox config
6. disable setr inventory:target true to false
7. working but lost target stash dumster

Want marker shop for this store remove target list but shop stop working if no target set to false

no error log in client or server side

![image](https://user-images.githubusercontent.com/4592618/207568668-5e802554-67e7-426e-8f27-eceee9143778.png)
![image](https://user-images.githubusercontent.com/4592618/207568759-81077d1a-2a52-4120-9e8c-b1b1765522ad.png)

We use last version of all version of ox (2 day ago)

Gregory-Hepicloud:
> that's how it works, yes

We want a workshop, with a marker system and some with a target system, but if only add locations, the shop does not work or if the target is displayed, but we want to type of shop is not possible ?

thelindat:
Doing some cleaning up and this is a copy of #960, so closing.

Issue #960, opened 2022-11-11
Use points alongside targets (e.g. shops)


Azakyel:
Add the possibility to choose for each shop if it use target system or not, it's a good feature for vending machine, but for others shops, the old system can be better

N-fire:
It's already the case, I think just don't put target coords.

Azakyel:
when i don't put target coords, the shops just didn't work .. maybe i miss something !

N-fire:
I might be wrong

juddisjudd:
Closing due to prolonged inactivity and grammatical errors.

PR #1073, opened 2022-12-31
Added peds to shops

TonybynMp4:
It's probably really badly implemented so don't hesitate to review with changes (like please do it)!
### know bug/problem:
You can't have targets AND peds for the same 'type' of shop
it won't spawn the peds because of the elseif (that's an easy fix buuuut....) 
if they did spawn (with the targets) then you can't open the ped's shops (i think the shop doesn't get added to the Shops table or something ? )
I tried changing the id's to i +100 when it's peds and it just never appears in the Shops table 🤷🏻


modules/shops/client.lua
```diff
@@ -106,6 +106,46 @@ client.shops = setmetatable(data('shops'), {
...
+
+							shops[id] = {
+								zoneId = shopid,
+								blip = blip and createBlip(blip, ped.loc)
+							}
+							local pedmodel = joaat(ped.ped)
+							RequestModel(pedmodel)
+							while not HasModelLoaded(pedmodel) do
+								Wait(0)
+							end
```
FjamZoo, line 123: There's a ox_lib function to do these makes it look cleaner

modules/shops/client.lua
```diff
@@ -106,6 +106,46 @@ client.shops = setmetatable(data('shops'), {
...
+							local PedId = CreatePed(0, ped.ped, ped.loc.x, ped.loc.y, ped.loc.z-1, ped.loc.w, true, true)
```
FjamZoo, line 124: Why is the peds here Networked? that means if multiple people join shops will be messed up

TonybynMp4: cause i'm dumb sorry x)

thelindat:
iirc, shops will be recreated when your "groups" change (jobs, gangs, whatever) which will lead to extra peds getting spawned in; also no handling for resource restart?

Really the handling for points and zones needs to be redone entirely, then peds added in; also requiring support for older format so people don't need to redo everything when updating.

TonybynMp4:
> iirc, shops will be recreated when your "groups" change (jobs, gangs, whatever) which will lead to extra peds getting spawned in;

indeed that definitely needs to be fixed somehow

> also no handling for resource restart?

Is it really necessary? the peds automatically go away (/ disappears it's kinda bugged i guess) and aren't targetable anymore, i don't think it really needs a restart function 🤔 

> Really the handling for points and zones needs to be redone entirely, then peds added in; also requiring support for older format so people don't need to redo everything when updating. This would ideally also drop support for qtarget as previously planned.

That sounds like a painful thing to do, i could probably replace the qtarget exports for ox_target to start if you want.

thelindat:
> > iirc, shops will be recreated when your "groups" change (jobs, gangs, whatever) which will lead to extra peds getting spawned in;
> 
> indeed that definitely needs to be fixed somehow

Rather than fixing this particular problem, rewriting everything would solve it.

> > also no handling for resource restart?
> 
> Is it really necessary? the peds automatically go away (/ disappears it's kinda bugged i guess) and aren't targetable anymore, i don't think it really needs a restart function 🤔

They'll stop being handled by the script and go do regular npc stuff; still weird and annoying.

> > Really the handling for points and zones needs to be redone entirely, then peds added in; also requiring support for older format so people don't need to redo everything when updating. This would ideally also drop support for qtarget as previously planned.
> 
> That sounds like a painful thing to do, i could probably replace the qtarget exports for ox_target to start if you want.

Yes, trying to upgrade things is a nuisance; more-so when you also need to keep compatibility.

TonybynMp4:
> Rather than fixing this particular problem, rewriting everything would solve it.

> They'll stop being handled by the script and go do regular npc stuff; still weird and annoying.

i think i figured out how to fix it temporarily, peds despawn at resource stop and when re-creating shops.
I wish i could help you rewrite the entire shop system but that's definitely outside of my capacities sadly :')

I also converted a few qtarget exports to ox_target.
I noticed the evidence and stashes would be annoying to convert so i'll try later today

thelindat:
I recommend waiting for the time being, as I said before
> the handling for points and zones needs to be redone entirely

This'll be a priority for the next release in order to resolve #422, #960, #1034

thelindat:
I've rewritten this crap nearly a dozen times now (wish I was exaggerating). I'm so unhappy with the overall structure for shops but it's not something I can just change because of backwards compatibility.

Probably going to have to be something to come with inventory v3, but my motivation to work on this stuff has been massively lacking for the past ~3 months.

TonybynMp4:
i think i fixed the issues there 🤔 

> This'll be a priority for the next release in order to resolve #422, #960, #1034

thelindat:
I’ve added ped spawning in without any of the other much-needed changes that shops need; mostly because backwards compatibility is too much of a hassle and making some temporary tweaks just to eventually overhaul things in a major update leaves a sour taste.

Peds are loaded alongside zones and use the same table/data, with the addition of adding ped and scenario fields.

Making zones work with markers is an issue because they conflict with the current setup, and people updating will suddenly find themselves loading a set of shops they don’t want.

PR #1092, opened 2023-01-14
fix (trunk/glovebox): CVehicle support of OX framework

ESK0:
Getting actual vehicle plate from CVehicle object in case fake plate is installed on vehicle and also prevent from opening trunk of different vehicle if you swap a plates on vehicles.

oxcore vehicle with fake plate installed:
**CVehicle**
image
**Plate**
image
**Inventory**
image

random vehicle:
**Plate**
image
**Inventory**
image


thelindat:
Not a good solution. Ideally vehicles will be stored by something other than the plate, but that requires some more framework dependent code and hasn't happened yet.

A callback event isn't really necessary either since the netid is currently being sent so everything can be handled server-side (in the future).

PR #1096, opened 2023-01-16
feat(crafting): labels and locales

Skryptific:
Adds dutch crafting locales and support for custom label and icon on crafting zones.
Also fixes custom labels for crafting benches as they are used in blips ,but not the inventory title.

### DIFF
--- a/modules/crafting/client.lua
+++ b/modules/crafting/client.lua
@@ -44,7 +44,7 @@ local function createCraftingBench(id, data)
 				zone.index = i
 				zone.options = {
 					{
-						label = 'Open Crafting Bench',
+						label = zone.label or 'Open Crafting Bench',
 						canInteract = data.groups and function()
 							return client.hasGroup(data.groups)
 						end or nil,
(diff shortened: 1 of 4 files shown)


modules/crafting/client.lua
```diff
@@ -44,15 +44,15 @@ local function createCraftingBench(id, data)
...
 				zone.options = {
 					{
-						label = 'Open Crafting Bench',
+						label = zone.label or locale('open_crafting_bench'),
 						canInteract = data.groups and function()
 							return client.hasGroup(data.groups)
 						end or nil,
 						onSelect = function()
-							client.openInventory('crafting', { id = id, index = i })
+							client.openInventory('crafting', { id = id, index = i, label = data.label })
```
thelindat, line 52: label?

thelindat (changes requested):

modules/crafting/client.lua
Skryptific: > label?

[label = data.label or locale('crafting_bench'),](https://github.com/overextended/ox_inventory/blob/6995bacbde584836541be7f97ebacd67c5d99760/client.lua#L136)

PR #1106, opened 2023-01-24
Update chinese simplified and chinese traditional files for 2.22.1

TomStilson:


locales/zh-tw.json
```diff
@@ -83,6 +83,7 @@
   "evidence_cannot_take": "階級不夠，無法拿取證物",
   "dumpster": "垃圾桶",
   "crafting_item": "製作 %s",
-  "crafting_bench": "工作台",
+  "crafting_bench": "工作臺",
```
FAr18, line 86:
```suggestion
  "crafting_bench": "工作台",
```

locales/zh-tw.json
```diff
@@ -83,6 +83,7 @@
...
+  "open_crafting_bench": "打開工作臺",
```
FAr18, line 87:
```suggestion
  "open_crafting_bench": "打開工作台",
```

FAr18 (changes requested):
"台" or "檯" should be used. And I think it's better to use fewer strokes between synonyms.

locales/zh-tw.json
TomStilson: Thank you for your advice, will update now.

PR #1117, opened 2023-01-31
fakeplate integration

xManKindx:
- allows for setup of fakeplate column in vehicle database. 
- checks both plate and fakeplate columns in database on trunk/glovebox inventory load to allow servers to attach fakeplates to vehicles and not lose inventory data.
- trim whitespace from beginning and end of plates less than 8 characters on inventory load. Spaces were causing issues with UpdateVehicle export
- ox/nd framework 'fakeColumn' set to 'fakeplate' but im not familiar with these 2 frameworks. can be set to false or changed if they have these fields already in their vehicle table

I made these edits purely for ESX fakeplate integration with ox_inventory. Sorry in advance if this is something unwanted

### DIFF
--- a/client.lua
+++ b/client.lua
@@ -549,7 +549,7 @@ local function registerCommands()
 					-- No storage or no glovebox
 					if (checkVehicle == 0 or checkVehicle == 2) or (not Vehicles.glovebox[vehicleClass] and not Vehicles.glovebox.models[vehicleHash]) then return end
 
-					local plate = GetVehicleNumberPlateText(vehicle)
+					local plate = string.strtrim(GetVehicleNumberPlateText(vehicle))
 					client.openInventory('glovebox', {id = 'glove'..plate, netid = NetworkGetNetworkIdFromEntity(vehicle) })
 
 					while true do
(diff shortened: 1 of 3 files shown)


thelindat:
Plates and "fake plates" are the most horribly designed least well thought out design pattern used in FiveM. ESX _depends_ on the plate for identification, and has no way to match a vehicle entity to the database except for its plate. If you change a vehicle's plate, you need to update the database plate - it shouldn't rely on some "fakeplate" column or necessitate weird workarounds in every _other_ script to work.

> trim whitespace from beginning and end of plates less than 8 characters on inventory load. Spaces were causing issues with UpdateVehicle export

I'll take another look at where trimming is needed. Such an annoying "feature" of ESX considering a plate is _always_ 8 string 🙄.

> ox/nd framework 'fakeColumn' set to 'fakeplate'

Ox uses an incrementing id column and also assigns a vin, plate isn't used as a primary identifier. God knows what nd is doing but I begged him not to use licenses as the primary identifier.

> I made these edits purely for ESX fakeplate integration

Basically enough reason for me _not_ to add it. Improvements need to be made for the goal of framework-agnostic code, not implementing weird workarounds/unofficial edits to 3rd party code.

PR #1124, opened 2023-02-07
Update fr.json

Own01:


dolutattoo:
`ui_remove_ammo` should be "Vider le chargeur"

Own01:
oups !
Thanks

PR #1144, opened 2023-02-13
Update weapon.lua | Add new and other weapons

ic3d00:
Update weapon.lua | Add new and other weapons


thelindat:
https://github.com/overextended/ox_inventory/tree/weapons

PR #1145, opened 2023-02-13
Add distance check config

GaluzaCZ:
Sometimes i need to open inv of other player that is far, so i think this will be cool to add


thelindat:
`/viewinv`

If you need to interact with it, tough; I'm not going to support disabling security.

PR #1173, opened 2023-03-01
qb-core functions update

xDope7137:
qbcore uses amount instead of count for item data


thelindat:
Would need this for any functions that return item data (GetItemBySlot, GetItemByName, GetItemsByName at a glance; and I think AddItem/RemoveItem?).

xDope7137:
> Would need this for any functions that return item data (GetItemBySlot, GetItemByName, GetItemsByName at a glance; and I think AddItem/RemoveItem?).

No, not really. Only functions which returns the amount/count of the items in a inventory is needed to be modified. GetItemBySlot might need to be edited, but none of the qb-core based scripts use that function anyways as its not really useful function in 99% of use cases. 

GetItemsByName is not used in qb-core I think. So GetItemByName had to be modified to support raw qb-core function.

xDope7137:
Tested the images on fresh qb-core server

modules/bridge/qb/server.lua
```diff
@@ -48,6 +48,12 @@ local function setupPlayer(Player)
 	Inventory.SetItem(Player.PlayerData.source, 'money', Player.PlayerData.money.cash)
 
 	QBCore.Functions.AddPlayerMethod(Player.PlayerData.source, "AddItem", function(item, amount, slot, info)
+		local info = info ~= nil and info or {}
+		local img = QBCore.Shared.Items[item] ~= nil and QBCore.Shared.Items[item].image or ''
+		if img == '' then
+			img = item..'.png'
+		end
+		info.image = string.gsub(img, ".png", "")
```
thelindat, line 56: Why? No.

modules/inventory/client.lua
```diff
@@ -13,8 +14,12 @@ function Inventory.OpenDumpster(entity)
...
 	end
 
-	if netId then
-		client.openInventory('dumpster', 'dumpster'..netId)
+	local dcoords = GetEntityCoords(entity)
+	local x = math.floor(dcoords.x)
+	local y = math.floor(dcoords.y)
+	dumpster = "dumpster|" .. x .. "|" .. y
+	if netId and dumpster then
+		client.openInventory('dumpster', 'dumpster'..dumpster)
```
thelindat, line 22: Avoid random unrelated changes.

server.lua
```diff
@@ -107,13 +107,7 @@ lib.callback.register('ox_inventory:openInventory', function(source, inv, data)
...
 
 			if not right then
-				local netid = tonumber(data:sub(9))
-
-				-- dumpsters do not work with entity lockdown. need to rewrite, but having to do
-				-- distance checks to some ~7000 dumpsters and freeze the entities isn't ideal
-				if netid and NetworkGetEntityFromNetworkId(netid) > 0 then
-					right = Inventory.Create(data, locale('dumpster'), inv, 15, 0, 100000, false)
-				end
+				right = Inventory.Create(data, locale('dumpster'), inv, 15, 0, 100000, false)
```
thelindat, line 110: ^

PR #1185, opened 2023-03-06
Use convar for inventory save interval

nathan-commits:


nathan-commits:
You're right, it's only useful for server crashes. Ten minutes is a bit too long for me, and I would find it very useful to be able to change this to 5 minutes using a convar. 

Can't wait for 3.0!

thelindat:
> it's only useful for server crashes

> I would find it very useful to be able to change this to 5 minutes

Surely you would want to solve server crashes, if that's something you're experiencing?

thelindat:
I still think it's stupid but if you want this merged, add some checks; I don't want weirdos setting the save interval super frequently (under 5 minutes even). This can also lead to duplication issues on server crash when player-inventory saves aren't handled internally (i.e. ESX).

eprimex:
Personally i wouldn't use this. I already have implemented a way on my server to make inventory and other resources save data on ESX save 1 player at once. Just needs a event that you make every resource listen to and pass only player server id and steam hex

PR #1203, opened 2023-03-14
feat(shops/server): item metadata weight

Skryptific:
When using metadata for weight in shops for example the clothing item, it now shows the correct weight in the shop instead of in your inventory after purchase.


Skryptific:
Added for crafting as well.

thelindat:
I'd classify both as fixes.

PR #1229, opened 2023-03-28
fear(server/setup): converting qb evidence

### LINKED ISSUE #1224: QB Police Evidence Conversion
**Is your feature request related to a problem? Please describe.**
locker numbers no longer have the items in them when changing evidence to use ox_inventory instead of qb-policejob

**Describe the solution you'd like**
Convert the police evidence stashes from qb db schema to ox table so that existing lockers contain the same items

**Describe alternatives you've considered**
N/A

**Additional context**
I'll probably PR this myself if someone else doesn't do it soon.

Manason:
fixes #1224

### DIFF
--- a/setup/convert.lua
+++ b/setup/convert.lua
@@ -135,6 +135,26 @@ local function ConvertESX()
 end
 
 local function ConvertQB()
+
+	local function convertQbItems(items)
+		local converted = {}
+		local slot = 0
+		for _, v in pairs(items) do
+			if Items(v?.name) and v.amount then
+				slot += 1
+				converted[slot] = {slot=slot, name=v.name, count=v.amount, metadata = type(v.info) == 'table' and v.info or {}}
+				if v.type == "weapon" then
+					converted[slot].metadata.durability = v.info.quality or 100
+					converted[slot].metadata.ammo = v.info.ammo or 0
+					converted[slot].metadata.components = {}
+					converted[slot].metadata.serial = GenerateSerial()
+					converted[slot].metadata.quality = nil
+				end
+			end
+		end
+		return converted
+	end
+
 	if started then
 		return warn('Data is already being converted, please wait..')
 	end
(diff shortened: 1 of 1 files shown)


setup/convert.lua
```diff
@@ -135,6 +135,26 @@ local function ConvertESX()
 end
 
 local function ConvertQB()
+
+	local function convertQbItems(items)
+		local converted = {}
+		local slot = 0
+		for _, v in pairs(items) do
+			if Items(v?.name) and v.amount then
```
Manason, line 143: I kept the v.amount check from looking at the other examples in this file, but I'm not sure if it's possible for a qb item to have no amount.

thelindat:
https://github.com/overextended/ox_inventory/blob/main/modules/bridge/qb/server.lua#L178

thelindat:
> > https://github.com/overextended/ox_inventory/blob/main/modules/bridge/qb/server.lua#L178
> 
> We should extract a common function to convert items that isn't tied to a player and then refactor the convert code to call the bridge. I may do that in a separate PR.

Player is technically optional here; and used only for the sake of syncing physical accounts to items (money).

Manason:
Calling the bridge conversion function to handle the conversion now.

setup/convert.lua
```diff
@@ -303,6 +303,46 @@ local function ConvertQB()
...
+					oxEvidence[name][#oxEvidence[name]+1] = items[k]
+				end
+			else
+				oxEvidence[name] = items
+			end
+		end
+
+		for name, items in pairs(oxEvidence) do
+			count += 1
+			parameters[count] = { "INSERT INTO ox_inventory (owner, name, data) VALUES ('', :name, :data) ON DUPLICATE KEY UPDATE name = :name, data = :data", {
```
thelindat, line 332:
```lua
"INSERT INTO ox_inventory (owner, name, data) VALUES ('', ?, ?) ON DUPLICATE KEY UPDATE name = VALUES(name), data = VALUES(data)"
```

thelindat (changes requested):

PR #1249, opened 2023-04-10
Multi-access inventories / inventory-sync

thelindat:
Allows for multiple players to access the same inventory, rather than the old behaviour of preventing additional players from opening an inventory.

inv.open always refer to the active inventory id (same as player id, if not in an inventory).
inv.openedBy is a set of player ids currently using the referenced inventory.

This changes behaviour of notifications somewhat to better reflect changes made to inventory state, but they may be somewhat spammy.

Currently only tested player-owned inventory when accessed by a single player, and standard stashes/vehicle stashes sync with two players. This cannot be merged until people have properly tested it and provided bug reports / feedback. There's almost definitely bugs causing desync at best, or item duplication at worst.

https://user-images.githubusercontent.com/65407488/230926091-c0033732-d293-48c9-9d62-6f6ae0a8a488.mp4

### DIFF
--- a/client.lua
+++ b/client.lua
@@ -1715,7 +1715,19 @@ RegisterNUICallback('buyItem', function(data, cb)
 			}
 		}, data[4])
 
-		SendNUIMessage({ action = 'refreshSlots', data = data[3] and {items = {{item = data[2]}, {item = data[3], inventory = 'shop'}}} or {items = {item = data[2]}}})
+		if data[3] then
+			SendNUIMessage({
+				action = 'refreshSlots',
+				data = {
+					items = {
+						{
+							item = data[3],
+							inventory = 'shop'
+						}
+					}
+				}
+			})
+		end
 	end
 
 	if message then
--- a/client.lua
+++ b/client.lua
@@ -853,20 +853,23 @@ local function updateInventory(items, weight)
 		client.setPlayerData('weight', weight)
 	else
 		for i = 1, #items do
-			local v = items[i].item
-			local item = PlayerData.inventory[v.slot]
+			if items[i].inventory == cache.serverId then
+				items[i].inventory = 'player'
+				local v = items[i].item
+				local item = PlayerData.inventory[v.slot]
 
-			if item?.name then
-				itemCount[item.name] = (itemCount[item.name] or 0) - item.count
+				if item?.name then
+					itemCount[item.name] = (itemCount[item.name] or 0) - item.count
+				end
+
+				if v.count then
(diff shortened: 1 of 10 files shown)


Manason:
Let's say player A attempts to give an item to player B, at the same time player B is transferring the same time from player A's inventory to their own. The give is received by the server first, but player B's transfer is not far behind and the sync has not yet occurred. By what mechanism is player B's transfer prevented?

thelindat:
> Let's say player A attempts to give an item to player B, at the same time player B is transferring the same item from player A's inventory to their own.

Players are still restricted from opening another player's inventory if that player is in any inventory.

> The give is received by the server first, but player B's transfer is not far behind and the sync has not yet occurred. By what mechanism is player B's transfer prevented?

If we ignore that you shouldn't be able to initiate swapslots and give at the same moment (due to my first point), the server has full authority here. Moving an item calls a nui callback, which triggers a server callback and awaits a response.

Data sent to the server is fromType and toType (all that matters if it's "player", which refers to their own inventory), fromSlot and toSlot which is the slot ids, and count (amount to move). The server handles all checks using minimal input from the client, and responds with the updated data for the referenced slots.

So player B's transfer is prevented if nothing exists in the referenced slot anymore, or they get a different item (or count) then they expected when initially moving the item.

https://github.com/overextended/ox_inventory/blob/a32c94796cc9f5659608b8eaeaa60bcb97417c43/modules/inventory/server.lua#L1365-L1660

Manason:
There's probably a very small chance of dupe. Unless you put a lock on accessing the inventory in memory, a 2nd read after the 1st read but before the 1st write will cause a dupe since the reading/writing is done in separate threads. Unless lua has some built-in sync behavior that two threads can't access the same piece of memory at the same time.

thelindat:
> There's probably a very small chance of dupe. Unless you put a lock on accessing the inventory in memory (or a lock on an individual slot before modifying using swap), a 2nd read after the 1st read but before the 1st write will cause a dupe since the reading/writing is done in separate threads.

There’s no multithreading so 1st read/write should always occur before 2nd except in the case somebody decided to yield in an event hook - but still need to test rather than theorise.

thelindat:
https://github.com/overextended/ox_inventory/pull/1249/commits/595b944574e5008f9d0ad9ac4291fefab68606ea should™ prevent any dupes from unexpected yielding; tested by adding `Wait(math.random(500, 2000))` to a swap hook. While swap A is pending, swap B, C, D will wait - fxserver bookmarking/scheduling should™ run each thread in the order it was created.

PR #1260, opened 2023-04-13
feat(inventory/propModel) add custom prop model support

mikigoalie:


thelindat:
Convar should be dropmodel or something.

PR #1266, opened 2023-04-16
feat(client): nested custom context buttons

Volumed:
Custom nested buttons. If you like buttons.

Use case: if you have a lot of interactions on an item and you want it to be neatly organized.

Schermafbeelding 2023-04-16 135519


Volumed:
I did it with groups because it was the least invasive. I would need to change more to how useButton functions with the id and index to make it work. I can still do it, if it's more desirable that way.

PR #1280, opened 2023-04-25
Discontinued: Allow admins to open any stash

JonasDev17:


thelindat:
Except I don't want that functionality and others may not, and this requires other changes to even implement.

JonasDev17:
> Except I don't want that functionality and others may not, and this requires other changes to even implement.

Thinking a bit more about this, you're right, this is not the way to go about it. What I actually wanted to implement is a way to open a stash with an event or function, bypassing these group checks. Would that be something you and others would want? @thelindat

thelindat:
I can do an export to open an inventory without the usual checks, but an event is just ripe for abuse; of course people will just do an event to call the export anyway but what can I do.

JonasDev17:
That'd be great! 

> but an event is just ripe for abuse; of course people will just do an event to call the export anyway but what can I do.

That's why I would've added an ace perms check but an export is all I need.

thelindat:
https://overextended.github.io/docs/ox_inventory/Functions/Server/forceOpenInventory/

PR #1282, opened 2023-04-26
Add trailers handle for inventory

Skulrag:
This PR add the possibility to open inventory when the player is positioned at the back of a trailer (all types).

Unfortunately, Cfx Natives does not provide a function IsThisModelATrailer(), but provide a function for all the others Model Types, so I had to add a full check to know if the model is a trailer or not.

After that added a code to check the entity bone name that is positioned at the back of trailers : wheel_rr (back right wheel).

Here is a quick code to show all the trailers bones indexes :

-- local boneNameListForTrailer = {
			-- 	'bodyshell',--  1
			-- 	'suspension_lm',--  1
			-- 	'suspension_lr',--  1
			-- 	'attach_male',-- 1
			-- 	'chassis_dummy',-- 1	//Center of the dummy  
			-- 	'wheel_lr',--	1//Wheel left, back  
			-- 	'wheel_rr'-- 1	//Wheel right, back  
			-- }

			-- for i=1, #boneNameList do
			-- 	boneIndex = GetEntityBoneIndexByName(entity, boneNameList[i])
			-- 	print('boneIndex', boneIndex)
			-- end


thelindat:
This would include some other vehicle types like submarinecar, and some trailers that shouldn't really have storage (i.e. anti-aircraft trailer). Might just be better to hardcode honestly.

Skulrag:
@thelindat I tried to send the entity on server-side and then check VehicleType but it always tells me the entity is not valid.

I'll do more research on how to get the entity server-side.

Else the only hope is FiveM adding a IsThisModelATrailer() func on client-side I guess ?

mikigoalie:
VehToNet from client and GetEntityFromNetworkId

Skulrag:
@thelindat The solution @mikigoalie proposed is working, I updated it on the PR. Hope you gonna like it 👌

thelindat:
Not perfect but it'll do for now.

PR #1295, opened 2023-04-30
feat(inventory/client): new export GetPlayerInventory

Manason:
Useful for integrating with scripts that have their own inventory UI such as drug processing tables, or player to player trade scripts

### DIFF
--- a/modules/inventory/client.lua
+++ b/modules/inventory/client.lua
@@ -41,6 +41,14 @@ else
 	Inventory.Dumpsters = dumpsters
 end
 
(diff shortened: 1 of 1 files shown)


FjamZoo:
Why not just use [GetInventory](https://overextended.github.io/docs/ox_inventory/Functions/Server/Inventory/GetInventory)?

Works with player inventories the same way that it works with stashes

thelindat:
Would rather call it GetPlayerItems for better accuracy.

PR #1298, opened 2023-05-01
at_images =/

MrNewb:
These may not fit styling but its been a while and I haven't seen any being added. I made all of these by taking screenshots and added a drop shadow.

these may kinda look like poo though..


MrNewb:
at_clip_drum_shotgun-no shadow
at_clip_drum_smg-no shadow
at_compensator-no shadow
at_scope_large-no shadow
 also did these without the backdrop and I lightened them up a bit in case they were to dark?

thelindat:
I've got some semi-automation setup for this stuff and just need to throw in model names; which I'd rather do than merge these for consistent lighting/style.

PR #1299, opened 2023-05-01
feat(crafting): crafting location not required

Manason:
- falls back to player position if both points and zones are nil
- untested


thelindat:
> feels confident merging without testing.

never have i ever

PR #1301, opened 2023-05-02
feat(bridge/qb) Added compatibility for the "GetItemByName" Function

NietThijmen:
This adds the "amount" field which is the same as "count" and "info" which is the same as "metadata"

### DIFF
--- a/modules/bridge/qb/server.lua
+++ b/modules/bridge/qb/server.lua
@@ -61,7 +61,7 @@ local function setupPlayer(Player)
 
 	QBCore.Functions.AddPlayerMethod(Player.PlayerData.source, "GetItemByName", function(item)
 		local search = Inventory.Search(Player.PlayerData.source, 'slots', item)
-		if not next(search) then
+		if next(search) then
 			return nil
 		end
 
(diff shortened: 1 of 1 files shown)


modules/bridge/qb/server.lua
```diff
@@ -60,7 +60,16 @@ local function setupPlayer(Player)
 	end)
 
 	QBCore.Functions.AddPlayerMethod(Player.PlayerData.source, "GetItemByName", function(item)
-		return Inventory.GetItem(Player.PlayerData.source, item, nil, false)
+		local search = exports.ox_inventory:Search(Player.PlayerData.source, 'slots', item)
+		if search == {} then
```
thelindat, line 64: Search will never be equal to {} (two different tables with their own hashes). You can just check `next(search)`.

modules/bridge/qb/server.lua
```diff
@@ -60,7 +60,16 @@ local function setupPlayer(Player)
...
+			return nil
+		end
+
+		local CurrentItem = search[1]
```
thelindat, line 68: lowercase?

thelindat:
https://github.com/overextended/ox_inventory/commit/5ea1c1731a0db19a981001ba04e569cd4e68f6d0

PR #1326, opened 2023-05-21
Getting correct plate from inventory type when inventory needs to be created

EpKouhia:
When inventory doesn't exist and it needs to be created inventory type being trunk or glovebox plate is only grab for type being "trunk". 

First commit should fix the problem where plate will be taken depending length of the type. 

Second commit adds check that configuration for that specific vehicle trunk of glovebox configuration exists in the data/vehicles.lua to prevent erroring out.

```lua
local inventoryId = 'glovebox'..plate
exports.ox_inventory:GetInventory(inventoryId)
exports.ox_inventory:AddItem(inventoryId, 'water', 20)
```

```
[ script:ox_inventory] plate:   box42FJH944
[ script:ox_inventory] [info] Failed to load vehicle inventory data (no entity exists with given plate).
```

### DIFF
--- a/modules/inventory/server.lua
+++ b/modules/inventory/server.lua
@@ -136,7 +136,7 @@ local function loadInventoryData(data, player)
 	end
 
 	if data.type == 'trunk' or data.type == 'glovebox' then
-		local plate = data.id:sub(6)
+		local plate = data.id:sub(data.type:len() + 1)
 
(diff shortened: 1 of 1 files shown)


EpKouhia:
Inventory ID will only consist "glove" or "trunk" where there is no need to check the type length to determine split.

PR #1343, opened 2023-06-01
Adding a bypassWeight parameters in metadata

lexinor:
If an item has bypassWeight in it's metadata the weight check is bypassed to help with containers.

Maybe someone has a better option, like to add a property directly in items


thelindat:
No.

PR #1372, opened 2023-06-19
crafting - Enable opening by export even outside zones or when the zone is not created at all.

polisek:
I needed to open crafting based on a specific condition and because I use target I always had to be near the zone and it was limited.

I added an unlockpos function to allow opening the crafting menu in other ways.

Example:
In a script where I have split illegal groups based on specific conditions and I couldn't use groupu so I needed this minor modification.

### DIFF
--- a/client.lua
+++ b/client.lua
@@ -175,7 +175,7 @@ function client.openInventory(inv, data)
 				return lib.notify({ id = 'cannot_perform', type = 'error', description = locale('cannot_perform') })
 			end
 
-			left = lib.callback.await('ox_inventory:openCraftingBench', 200, data.id, data.index)
+			left = lib.callback.await('ox_inventory:openCraftingBench', 200, data.id, data.index, data.unlockpos)
 
 			if left then
 				right = CraftingBenches[data.id]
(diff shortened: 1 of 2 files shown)


thelindat:
or just, like, don't define any coords for the bench.

polisek:
I'd like that, but if I don't fill in the zone I get that data.zones on the client is equal to nil.

But if I put the condition in modules/crafting/client.lua
if data.zones then
Then crafting won't open because there is a check from the callback on the bench.

server side
openCrafingBench
is if bench then
if #(GetEntityCoords(GetPlayerPed(source)) - coords) > 10 then return end


This is the reason why this PR was created .

polisek:
> or just, like, don't define any coords for the bench.

I'd like that, but if I don't fill in the zone I get that data.zones on the client is equal to nil.

But if I put the condition in modules/crafting/client.lua
if data.zones then
Then crafting won't open because there is a check from the callback on the bench.

server side
openCrafingBench
is if bench then
if #(GetEntityCoords(GetPlayerPed(source)) - coords) > 10 then return end


This is the reason why this PR was created in the first place.

PR #1410, opened 2023-07-19
tweak(client): only show give item playerlist if more than 1 player is nearby

jellyton69:
If there is only one player nearby, we should just give the player the item instead of asking which player to give the item to.


thelindat:
Rather than wrapping the existing code in an if statement, rather just add the new check to the top, i.e.

```lua
local playerCount = #nearbyPlayers

if playerCount == 0 then return end

if playerCount == 1 then
    ...
end
```

PR #1412, opened 2023-07-22
feat(client): specify displayMetadata order

Manason:
- added optional parameter to displayMetadata to specify an order of metadata entries to display. Made some UI code adjustments to change to an array to preserve order.

### DIFF
--- a/modules/items/client.lua
+++ b/modules/items/client.lua
@@ -2,18 +2,27 @@ if not lib then return end
 
 local Items = require 'modules.items.shared' --[[@as table<string, OxClientItem>]]
 
----@param metadata table|string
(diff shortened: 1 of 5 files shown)


thelindat:
Why order as a new array/parameter and not just accept an array for metadata?

```lua
displayMetadata({
    { a = 'A' }, -- show first
    { b = 'B' }, -- show second
})
```

Manason:
> Why order as a new array/parameter and not just accept an array for metadata?
> 
> ```lua
> displayMetadata({
>     { a = 'A' }, -- show first
>     { b = 'B' }, -- show second
> })
> ```

done

PR #1415, opened 2023-07-24
fix(modules/items): durability nil value

ItzMartijn:
When setting an items' durability with SetDurability, would not set the durability when durability has not been defined yet within the metadata


thelindat:
`or value` makes more sense than defaulting to 100.

PR #1424, opened 2023-08-02
fix(inventory/server): CanCarryItem bug when weight is nil instead of 0

TheiLLeniumStudios:


thelindat:
Weight should never be nil in the first place, so would be good to know how that's happening instead.

TheiLLeniumStudios:
@thelindat so I just checked and its because some items do not have the `weight` key in the items table which causes the weight to show up as nil. I haven't had a chance to look at how defaults are handled for items but I'd assume it does automatically set that to 0? Or is it something that should be required for every single item that is defined?

thelindat:
> I'd assume it does automatically set that to 0

https://github.com/overextended/ox_inventory/blob/main/modules/items/shared.lua#L12

TheiLLeniumStudios:
> 

Hmm, weird. I will debug the resource that causes this. I can't even be the item not existing in data/items.lua because because otherwise it would just not go further into the function. For reference, a stack trace (Not very helpful for you but just shows that somehow it does trigger the error):

```
[ script:ox_inventory] SCRIPT ERROR: @ox_inventory/modules/inventory/server.lua:1359: attempt to perform arithmetic on a nil value (local 'weight')
[ script:ox_inventory] > ref (@ox_inventory/modules/inventory/server.lua:1359)
[ script:ox_inventory] > handler (@apollo_one/server/recyclejob.lua:13)
[   script:apollo_one] SCRIPT ERROR: @apollo_one/server/recyclejob.lua:13: 
[   script:apollo_one]  An error occurred while calling export `CanCarryItem` in resource `ox_inventory`:
[   script:apollo_one] 
[   script:apollo_one]  ---
```

TheiLLeniumStudios:
okay this is strange. I haven't been able to reproduce this at all now. Apologies for just wasting your time then. Will close as this is not needed

PR #1431, opened 2023-08-05
feat(client): adding support for plate statebag

itIsMaku:
- adding support for license plate removal scripts (reading license plate from statebag if the plate is empty) for trunks and gloveboxes


thelindat:
This is kind of weird and making assumptions about how other scripts work (I assume based off whatever you use).
Should instead be a function that defaults to returning the native plate; people can just edit it from there.

Of course, nobody will use it (see all the complaints about shops taking money only).

PR #1432, opened 2023-08-06
fix(server): Save inventories on restart

SamShanks1:
Previously just before scheduled server restarts you were able to take an item out of your stash into your inventory and then when re-joining the item would be in your inventory and stash duplicating. Changing this prevented it.

### DIFF
--- a/modules/inventory/server.lua
+++ b/modules/inventory/server.lua
@@ -2255,6 +2255,15 @@ AddEventHandler('txAdmin:events:serverShuttingDown', function()
 	Inventory.SaveInventories(true)
 end)
 
+AddEventHandler('txAdmin:events:scheduledRestart', function(eventData)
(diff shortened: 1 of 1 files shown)


thelindat:
This is what the `txAdmin:events:serverShuttingDown` event is for; so if it's somehow having issues it needs to be looked at. This would also affect other scripts that save with that event too (e.g. ESX) and I've not heard such complaints.

SamShanks1:
My bad, I thought that was only to do with unscheduled restarts (I misread the docs). I guess it could be the timings that affect it. 
I believe it was happening on a QB server, someone just messaged me stating how they did it and that fixed it weirdly. Im gonna get a video and test it myself on a base QB and OX server to see if its something that was different on their server.

Ill close this and then reopen it if I find anything interesting

PR #1442, opened 2023-08-13
Added locales server.lua

nemaDev:
New lines were added for server side commands and in es.json a missing line was added.


thelindat:
These locale keys are certainly something. Not sure why you did 3 PRs too.

PR #1465, opened 2023-08-27
Create drop on ground

BrunoSollar:
Previously, when the player dropped the item, the marker would float. With this commit, when the player drops the item, the marker is on the ground.
image


thelindat:
Why would I want it on the ground?

BrunoSollar:
> Why would I want it on the ground?

Having the marker on the ground when the player drops an item provides a more realistic in-game experience.

PR #1473, opened 2023-09-04
Marginal durability stacking feature

Repentz:
https://imgur.com/CCMPWNe

Upon swapping in inventory and attempting to stack an item (stack logic):

- It first extracts the 'durability' values of the two items. If the durability does not exist, it defaults to 0.
- It then clones the metadata of the two items for comparison, excluding the 'durability' property.
- It calculates the percentage difference between the two durabilities.
- It then checks if the items' names are the same and the rest of their metadata match.
- If the above check passes, it further checks if the difference in durability is 20% or less.
- If it is, it concludes that their durabilities are close enough. It then sets the durability of both items to the lower of the two durabilities.
- If it's more than 20%, it informs that their durabilities are too different & that it will proceed to swap logic (not provided in this script).

### DIFF
--- a/modules/inventory/server.lua
+++ b/modules/inventory/server.lua
@@ -1692,8 +1692,32 @@ lib.callback.register('ox_inventory:swapItems', function(source, data)
 					toData, fromData = Inventory.SwapSlots(fromInventory, toInventory, data.fromSlot, data.toSlot)
 				end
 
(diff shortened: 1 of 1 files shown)


Repentz:
multiple duplication bugs found, will investigate.

PR #1474, opened 2023-09-04
feat(crafting): Adds export to create new crafting

BeastPT:
Add possibility to create new crafting with started resource and open with export openInventory


MrLutin (approved):
Yes! this is what y wanted to make by myself ! Nice job !

thelindat:
Crafting is pending a partial rework so I'm not merging stuff like this.

PR #1497, opened 2023-10-03
feat(rename) Add rename button

NikXpro:


FilipeCuco:
why?

NikXpro:
This is very useful for renaming an object, for example, outfits or any item with metadata that appears multiple times in the inventory. It's highly requested on many servers.

Vdxxxxx:
Lourd de oufff

PR #1508, opened 2023-10-18
fix(mysql): db fix for ConfiscateInventory

xViperAG:


xViperAG:
Realized there's an error with the saving, going to revert and see what I can do to fix.

PR #1527, opened 2023-11-16
update(all/locales): unify translate variable, add missing CS locale

mikigoalie:
Was looking for missing locales on weapons and stumbled upon this. I haven't found any other use case other than mostly UI, so I unified the prefix and added some missing CS locales.


mikigoalie:
Would appreciate a feedback or reason so I can do better PR next time.

Ammo_type seems to be used only in ui, nowhere else, unifying it with other variables seems logical.

PR #1530, opened 2023-11-19
feat: register external items

Andyyy7666:
1. Create a resource:
2. create `test.lua` and add this.
```lua
return {
	['testburger'] = {
		label = 'Test Burger',
		weight = 220,
		degrade = 60,
		client = {
			status = { hunger = 200000 },
			anim = 'eating',
			prop = 'burger',
			usetime = 2500,
			-- export = 'ox_inventory_examples.testburger'
		},
		server = {
			-- export = 'ox_inventory_examples.testburger',
			test = 'what an amazingly delicious burger, amirite?'
		},
		buttons = {
			{
				label = 'Lick it',
				action = function(slot)
					print('You licked the burger')
				end
			},
			{
				label = 'Squeeze it',
				action = function(slot)
					print('You squeezed the burger :(')
				end
			}
		},
		consume = 0.3
	},
}
```

3. add to `fxmanifest.lua`:
```lua
file "test.lua"
ox_inventory_items "test.lua"
```

Done, this is what the community has asked for. Doesn't have to be over complicated this is perfect, only thing is that the resource that has the items has to be started before ox inventory which means that if a resource has ox inventory as a dependecy can't use this. I personally don't usually have ox_inventory as a dependency but have other ways to check.

### DIFF
--- a/modules/items/shared.lua
+++ b/modules/items/shared.lua
@@ -55,6 +55,55 @@ local function newItem(data)
 	ItemList[data.name] = data
 end
 
+---@param resourceName string
+---@return table?
+local function getExternalItems(resourceName)
+    local num = GetNumResourceMetadata(resourceName, "ox_inventory_items")
+    if num < 1 then return end
+    for i=0, num-1 do
+        local file = GetResourceMetadata(resourceName, "ox_inventory_items", i)
+        local path = ("@%s.%s"):format(resourceName, file:gsub(".lua", ""):gsub("/", "."))
+        local items = require(path)
+        if not items or type(items) ~= "table" then
+            return lib.print.warn(("Items from resource (%s) unable to load!"):format(resourceName))
+        end
+        return items
+    end
+end
+
+local function registerExternalItems()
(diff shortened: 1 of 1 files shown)


Andyyy7666:
The check for the external resource to be started can be removed and then ox inventory can be a dependency because it will also read items from stopped resources.

jag3dagster:
Local iteration on a resource isn't handled in this code. Say a developer was working on their resource and needed to iterate on an item for some reason, they would need to restart their resource and then restart ox_inventory. Not devistating, but definitely annoying.

Andyyy7666:
Yes I agree, that wouldn't be a problem if this below wasn't the case.
> not as simple as just adding an export and requires rewriting large parts of the code and the UI to properly support it. - linden

It's annoying but `ox_inventory/data/items.lua` can be used for development. This is just useful for when releasing resources to the public.

thelindat:
I don't want 3rd party files loaded as modules or into the inventory _ENV.

PR #1548, opened 2023-12-10
shop society

DARKAOFF:
Adding an optional company account to fund a store account


thelindat:
This adds a strict ESX dependency, which is obviously no good. I also don't speak French.

PR #1553, opened 2023-12-13
update(client): fire event on mounting weapon components

mikigoalie:


thelindat:
Technically can just use ox_inventory:updateInventory or ox_inventory:currentWeapon events fwiw

mikigoalie:
I will check and test again, but these events get triggered on reloads as well.. Not too convenient imho. But will double check

thelindat:
First triggers on any update event. Second when the current weapon or its metadata changes; which gives you what you need.

mikigoalie:
Fair enough, just thought it'd be little more convenient for easier attachment handling, but feel free to close it. Mentioned events are sufficient for my specific needs

client.lua
```diff
@@ -1542,6 +1543,7 @@ RegisterNUICallback('removeComponent', function(data, cb)
 
 					if success then
 						RemoveWeaponComponentFromPed(playerPed, currentWeapon.hash, component)
+						TriggerEvent('ox_inventory:updateWeaponComponents', "removed", currentWeapon, component)
```
thelindat, line 1546: `ox_inventory:updateWeaponComponent`.
Use `'` rather than `"` (i.e. 'removed', 'added').
Send the item name of the component being used (not just the hash).
Don't think there's much need to send currentWeapon at all.

mikigoalie: Done in latest commit, but double check it. I'll also PR 3095 weapons and components later

thelindat:
```lua
TriggerEvent('ox_inventory:updateWeaponComponent', action --[['removed' | 'added']], component --[[hash]], componentItem --[[string]])
```

PR #1555, opened 2023-12-13
update module/shop/client.lua 

DARKAOFF:
small modification which puts point.invId as a practical option when, like me, you want to create random NPCs for black markets


DARKAOFF:
only him to take into consideration
[43d9c76](https://github.com/overextended/ox_inventory/pull/1555/commits/43d9c7685b70545d6aeb942d8a33b7bc31b85d22)

PR #1559, opened 2023-12-17
fix: nd license check

Andyyy7666:
nd v1 had `.data` as metadata in v2 it's `.metadata` and there's a function to get specific metadata


alexcris9195:
if you're in the car you can't give a item to another player.

PR #1562, opened 2023-12-19
add CanCarryItems for multiple items

Andyyy7666:
New function:
```lua
ox_inventory:CanCarryItems(source, {
    ["weapon_pistol"] = 1,
    ["ammo-9"] = 30
})
```

Kept old function:
```lua
ox_inventory:CanCarryItem(source, "weapon_pistol", 1)
```

### DIFF
--- a/modules/inventory/server.lua
+++ b/modules/inventory/server.lua
@@ -1347,34 +1347,32 @@ exports('RemoveItem', Inventory.RemoveItem)
 ---@param count number
 ---@param metadata? table | string
 function Inventory.CanCarryItem(inv, item, count, metadata)
+    local totalItems = type(item) == "table" and item or {[item]=count}
 	if type(item) ~= 'table' then item = Items(item) end
+    inv = Inventory(inv) --[[@as OxInventory]]
 
-	if item then
-		inv = Inventory(inv) --[[@as OxInventory]]
+    local totalWeight = 0
+    for itemName, itemCount in pairs(totalItems) do
+        local item = Items(itemName)
+        if not item then goto skip end
+        
+        local itemSlots, _, emptySlots = Inventory.GetItemSlots(inv, item, type(metadata) == 'table' and metadata or { type = metadata or nil })
+        if not itemSlots or not next(itemSlots) and emptySlots == 0 or not item.stack and emptySlots < itemCount then return end
 
-		if inv then
-			local itemSlots, _, emptySlots = Inventory.GetItemSlots(inv, item, type(metadata) == 'table' and metadata or { type = metadata or nil })
+        local weight = metadata and metadata.weight or item.weight
+        if weight == 0 then goto skip end
 
-			if not itemSlots then return end
+        local itemsWeight = (weight * itemCount)
+        local newWeight = inv.weight + 
+        totalWeight += itemsWeight
+        if newWeight > inv.maxWeight or totalWeight > inv.maxWeight then
+            TriggerClientEvent('ox_lib:notify', inv.id, { type = 'error', description = locale('cannot_carry') })
+            return false
+        end
 
-			local weight = metadata and metadata.weight or item.weight
-
-			if next(itemSlots) or emptySlots > 0 then
-				if not count then count = 1 end
-				if not item.stack and emptySlots < count then return false end
-				if weight == 0 then return true end
-
-				local newWeight = inv.weight + (weight * count)
-
-				if newWeight > inv.maxWeight then
-					TriggerClientEvent('ox_lib:notify', inv.id, { type = 'error', description = locale('cannot_carry') })
-					return false
-				end
-
-				return true
-			end
-		end
-	end
+        ::skip::
+    end
+    return true
 end
 exports('CanCarryItem', Inventory.CanCarryItem)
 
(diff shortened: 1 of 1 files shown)


modules/inventory/server.lua
```diff
@@ -1342,6 +1342,37 @@ function Inventory.RemoveItem(inv, item, count, metadata, slot, ignoreTotal)
 end
 exports('RemoveItem', Inventory.RemoveItem)
 
+---@param inv inventory
+---@param items table
```
thelindat, line 1346: `---@param items table<string, number>` would be more accurate, however it would make more sense for both count _and_ metadata to be tied to the item. The current structure also wouldn't support adding two items with the same name, but separate metadata.

This would be better.

```lua
exports.ox_inventory:CanCarryItems(invId, {
   { 'bread', 1, { durability = 0 } },
   { 'bread', 1, { durability = 100 } },
})
```

modules/inventory/server.lua
```diff
@@ -1342,6 +1342,37 @@ function Inventory.RemoveItem(inv, item, count, metadata, slot, ignoreTotal)
...
+---@param metadata? table | string
+function Inventory.CanCarryItems(inv, items, metadata)
+    inv = Inventory(inv) --[[@as OxInventory]]
+
+    local totalWeight = 0
+    for itemName, itemCount in pairs(totalItems) do
```
thelindat, line 1352: Line break between local and for.

modules/inventory/server.lua
```diff
@@ -1342,6 +1342,37 @@ function Inventory.RemoveItem(inv, item, count, metadata, slot, ignoreTotal)
...
+        local item = Items(itemName)
+        if not item then goto skip end
```
thelindat, line 1354: Line break.

modules/inventory/server.lua
```diff
@@ -1342,6 +1342,37 @@ function Inventory.RemoveItem(inv, item, count, metadata, slot, ignoreTotal)
...
+        
+        local itemSlots, _, emptySlots = Inventory.GetItemSlots(inv, item, type(metadata) == 'table' and metadata or { type = metadata or nil })
+        if not itemSlots or not next(itemSlots) and emptySlots == 0 or not item.stack and emptySlots < itemCount then return end
+
+        local weight = metadata and metadata.weight or item.weight
+        if weight == 0 then goto skip end
```
thelindat, line 1360: Line break.

modules/inventory/server.lua
```diff
@@ -1342,6 +1342,37 @@ function Inventory.RemoveItem(inv, item, count, metadata, slot, ignoreTotal)
...
+
+        local itemsWeight = (weight * itemCount)
+        local newWeight = inv.weight + 
+        totalWeight += itemsWeight
```
thelindat, line 1364: huh

modules/inventory/server.lua
```diff
@@ -1342,6 +1342,37 @@ function Inventory.RemoveItem(inv, item, count, metadata, slot, ignoreTotal)
...
+        if newWeight > inv.maxWeight or totalWeight > inv.maxWeight then
+            TriggerClientEvent('ox_lib:notify', inv.id, { type = 'error', description = locale('cannot_carry') })
+            return false
+        end
+
+        ::skip::
+    end
+    return true
```
thelindat, line 1372: Line break.

modules/inventory/server.lua
```diff
@@ -1342,6 +1342,37 @@ function Inventory.RemoveItem(inv, item, count, metadata, slot, ignoreTotal)
...
+end
+exports('CanCarryItems', Inventory.CanCarryItems)
```
thelindat, line 1374: Line break.

modules/inventory/server.lua
```diff
@@ -1342,6 +1342,37 @@ function Inventory.RemoveItem(inv, item, count, metadata, slot, ignoreTotal)
...
+---@param metadata? table | string
+function Inventory.CanCarryItems(inv, items, metadata)
+    inv = Inventory(inv) --[[@as OxInventory]]
+
+    local totalWeight = 0
+    for itemName, itemCount in pairs(totalItems) do
+        local item = Items(itemName)
+        if not item then goto skip end
+        
+        local itemSlots, _, emptySlots = Inventory.GetItemSlots(inv, item, type(metadata) == 'table' and metadata or { type = metadata or nil })
```
thelindat, line 1356: Would be preferable to iterate over the inventory once, rather than per item. (requires new code)

thelindat:
Doesn't handle cases where there are limited slots available?

i.e. I want to add 3 items - there is 1 empty slot and 1 of the items will stack, making the last item "uncarryable".

modules/inventory/server.lua
Andyyy7666: Why do you want metadata to check if a player can carry an item, I think simply just checking if they have enough weight and available slots should be enough.

modules/inventory/server.lua
Andyyy7666: Should’ve been `local newWeight = inv.weight + itemsWeight` must’ve accidentally removed that.

modules/inventory/server.lua
thelindat: 1. Items with different metadata cannot stack with eachother.
2. Metadata can adjust item weight (consider a water bottle that empties overtime).

Issue #1466, opened 2023-08-28
Set lifetime on custom drops


VanishedMC:
**Is your feature request related to a problem? Please describe.**
All temporary inventories currently have a lifetime of 10 minutes, it would be great if this could instead be passed as an optional argument to `exports.ox_inventory:CustomDrop` and `exports.ox_inventory:CreateTemporaryStash`

**Describe the solution you'd like**
A new optional parameter for these 2 exports

**Describe alternatives you've considered**
Only alternative I currently have is changing the global 10 minute life time to some other value, though this isn't nearly as nice of a solution

PR #1571, opened 2023-12-23
Feat(inventory): Temp stash and custom drop custom time

FjamZoo:
This will close: https://github.com/overextended/ox_inventory/issues/1466

If you know a better method let me know 👍

### DIFF
--- a/modules/inventory/server.lua
+++ b/modules/inventory/server.lua
@@ -1455,7 +1455,7 @@ local function CustomDrop(prefix, items, coords, slots, maxWeight, instance, mod
 	inventory.items, inventory.weight = generateItems(inventory, 'drop', items)
 	inventory.coords = coords
 
-    if time then
+    if clearTime then
 		inventory.time = os.time() - clearTime * 60
 	end
     
--- a/modules/inventory/server.lua
+++ b/modules/inventory/server.lua
@@ -1456,7 +1456,7 @@ local function CustomDrop(prefix, items, coords, slots, maxWeight, instance, mod
 	inventory.coords = coords
 
     if time then
-		inventory.time = time
+		inventory.time = os.time() - time * 60
 	end
     
 	Inventory.Drops[dropId] = {
@@ -2666,7 +2666,7 @@ function Inventory.CreateTemporaryStash(properties)
 	inventory.coords = coords
 
 	if properties.time then
-		inventory.time = properties.time
+		inventory.time = os.time() - properties.time * 60
 	end
 
 	return inventory.id
--- a/modules/inventory/server.lua
+++ b/modules/inventory/server.lua
@@ -1446,7 +1446,7 @@ local function generateInvId(prefix)
 	end
 end
 
-local function CustomDrop(prefix, items, coords, slots, maxWeight, instance, model)
+local function CustomDrop(prefix, items, coords, slots, maxWeight, instance, model, time)
 	local dropId = generateInvId()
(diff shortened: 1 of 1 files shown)


FjamZoo:
for use case something like

```lua
os.time() - 60 * 10  -- Stashes will now last 10 minutes extra
```

kinda shit implementation but no other good way I feel like

thelindat:
- Would make more sense to pass the number of minutes you want it to last, rather than a timestamp.
- Still only a rough lifetime as it depends on the save frequency.
- Do people want it to timeout x time after it was last used, or after x time?

People could just make their own timeout/wait/loop/cronjob in their scripts and handle removal themselves using `RemoveInventory`.

FjamZoo:
Yeah I feel like passing a time they want it to last would make more sense, would just require me to move inventoryClearTime a little up and then do calculations but it's still gonna be kinda ugly

Would look something like this
```lua
local function CustomDrop(prefix, items, coords, slots, maxWeight, instance, model, time)
	local dropId = generateInvId()
	local inventory = Inventory.Create(dropId, ('%s %s'):format(prefix, dropId:gsub('%D', '')), 'drop', slots or shared.playerslots, 0, maxWeight or shared.playerweight, false, {})

	if not inventory then return end

	inventory.items, inventory.weight = generateItems(inventory, 'drop', items)
	inventory.coords = coords

	if time then
		inventory.time = os.time() - (time * 60) - inventoryClearTime
	end

	Inventory.Drops[dropId] = {
		coords = inventory.coords,
		instance = instance,
		model = model,
	}

	TriggerClientEvent('ox_inventory:createDrop', -1, dropId, Inventory.Drops[dropId])

    return dropId
end
```

Or as you mentioned leave inventoryClearTime out and be a rough timeout instead of an absolote

FjamZoo:
I made it a rough lifetime instead, easiest solution for now and worse case the inventory will just last a little longer

thelindat:
> Do people want it to timeout x time after it was last used, or after x time?
>
> People could just make their own timeout/wait/loop/cronjob in their scripts and handle removal themselves using RemoveInventory (we just need an option to keep it alive).

Covers more use-cases. Could just have a negative value mean it doesn't get removed during save loop. Rather than just "time", something like "clearTime" to be more descriptive.

FjamZoo:
> Covers more use-cases. Could just have a negative value mean it doesn't get removed during save loop.
If u want to switch to something like that I am all for it, just trying to get rid of some of these "suggestions".

I did do everything else u asked so lmk which route u want to go for.

modules/inventory/server.lua
```diff
@@ -2660,6 +2665,10 @@ function Inventory.CreateTemporaryStash(properties)
 	inventory.items, inventory.weight = generateItems(inventory, 'drop', properties.items)
 	inventory.coords = coords
 
+	if properties.time then
```
thelindat, line 2668: `properties.clearTime`

thelindat:
This doesn't even work right? inv.time gets updated everytime an inventory is accessed.

mikigoalie:
> This doesn't even work right? inv.time gets updated everytime an inventory is accessed.

idk, but me likey the idea of setting time for custom stashes.. Perhaps time in minutes for it to disappear since last interaction and false / true / 0 for it to stay existing until restart or whatever? I'm working on a house robbery resource where I'm using temp stashes, if you want, I can rewrite this whole logic and make new pr

PR #1581, opened 2024-01-01
feat: battlerifle & snow launcher

mikigoalie:


mikigoalie:
requires b3095, not stable yet, therefore no need to merge now. Weapon "hacking device" is not included intentionally. Had this commit drafted for myself so i sent pr now, Happy nye

Issue #1582, opened 2024-01-01
Cant use Renewed Phone


jessemel:
**Describe the bug**
When I try to use the phone it won't open and I get a list of errors. 
I have tried to go through and follow the errors and find changes and nothing works. I am starting a second server so this server is brand new. I just installed ox_inventory then installed the renewed-phone qb-phone and this is where I am. Any advice on what I should try next?

I have done the conversion and everything else seems to work fine.

Phone I'm using
https://github.com/Renewed-Scripts/qb-phone

**Framework**
Using QBCore

**Resource version**
'cerulean' - most recent release

**To Reproduce**
1. Fly in
2. Hit "M" and it won't open an errors in F8 appear

**Expected behavior**
Phone should open

**Screenshots**
![image](https://github.com/overextended/ox_inventory/assets/131414893/38a76430-421a-4407-9a7e-a04d4ea8f5b7)

**Additional context**
![image](https://github.com/overextended/ox_inventory/assets/131414893/2446efda-d134-4ad5-a748-67e50f0d245f)
![image](https://github.com/overextended/ox_inventory/assets/131414893/0d1c2b92-9c1b-4453-adad-9d7d59810932)
![image](https://github.com/overextended/ox_inventory/assets/131414893/4e801e72-6baf-4376-9c73-bc01a83ddf77)
![image](https://github.com/overextended/ox_inventory/assets/131414893/cc9ae0ef-3cfb-4828-b2f1-76e4e1ad5871)

PR #1587, opened 2024-01-03
fix(bridge/qb): "attempt to compare number with boolean" error fix

someone005:
Fix of issue **[1582](https://github.com/overextended/ox_inventory/issues/1582)**

### DIFF
--- a/modules/bridge/qb/client.lua
+++ b/modules/bridge/qb/client.lua
@@ -61,19 +61,17 @@ end
 local function hasItem(items, amount)
     amount = amount or 1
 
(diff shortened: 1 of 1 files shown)


thelindat:
Going to assume this happens when checking for an item that does not exist? Your fix isn't really appropriate though.

thelindat:
> it would fix this error because GetItemCount always returns number instead of boolean so there is no possibility for this error to happen again

You didn't answer my question, and you can just do "or 0" so that you don't end up comparing false to a number.

PR #1601, opened 2024-01-18
Re-sort photos into folders

bicatcho11:
The modification will look more organized and will be considered a radical change by users, but I think everyone will like it


LukeWasTakenn:
this is a breaking change.

bicatcho11:
@LukeWasTakenn why close
I wanted to fix almost all the bugs
When you want to release a new version
You can use my withdrawal
can reopen ?

PR #1614, opened 2024-02-02
feat(client/weapons): add useSlot weapon anim skip

Demigod916:
this adds a new optional param to the useSlot export, it allows for you to disable the anim for weapons if enabled when calling useSlot externally. while this is a little bit niche, i think this can come in handy in any instances where you need to set a weapon to a player immediately.

### DIFF
--- a/client.lua
+++ b/client.lua
@@ -337,7 +337,7 @@ local Items = require 'modules.items.client'
 local usingItem = false
 
 ---@param data { name: string, label: string, count: number, slot: number, metadata: table<string, any>, weight: number }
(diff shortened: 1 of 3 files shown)


thelindat:
A generic skipAnim or such would probably make more sense (though anims are tied to the progress bar).

Demigod916:
I'll work on a pr for a generic skipAnim then. Thanks

Demigod916:
ive updated the PR to now be more of a generic anim skip. so if you manually use the useSlot export and pass true for 2nd arg, it will skip the weapon anim and then if the item is not a weapon it will skip the progress/anim for it instead

PR #1631, opened 2024-02-26
fix inventory:weight Maximum carry capacity override in qbcore framwork

CodeRed57:
check players inventory if can carry before adding items


0xDEMXN:
I may be wrong but it's intended behavior and it's up to you to check for weight before adding an item.
If that isn't the case you shouldn't do this in bridge but in the inventory itself.

Issue #1634, opened 2024-03-01
:GetCurrentWeapon Conflict


googleoblivion:
https://github.com/overextended/ox_inventory/blob/9b7680451c602d1632e797fb454688dc62ed41f4/modules/weapon/client.lua#L98

We don't trigger the updateWeapon server event here on disarm, which is where the behavior to call inv.weapon = nil is. Which leads to caching the weapon and not allowing the function to get the actual current weapon. You can fix with changing this check to

`if not source then`

but this may lead to breaking a security measure that I don't understand is occuring here. I know this is checking for if it's the server calling the function, but I can't also find anywhere the server is calling the function to client.

Randolio:
Adding on to this, I was also confused. For myself, I modified the GetCurrentWeapon function temporarily to:

```
function Inventory.GetCurrentWeapon(inv)
	local ped = GetPlayerPed(inv)

	if GetSelectedPedWeapon(ped) == `WEAPON_UNARMED` then
		return nil 
	end

	inv = Inventory(inv) --[[@as OxInventory]]

	if inv?.player then
		local weapon = inv.items[inv.weapon]

		if weapon and Items(weapon.name).weapon then
			return weapon
		end

		inv.weapon = nil
	end
end
exports('GetCurrentWeapon', Inventory.GetCurrentWeapon)
```

But that still felt like an unnecessary budget work around.

PR #1635, opened 2024-03-01
fix(weapon/client): updateWeapon source check

googleoblivion:
PR fix for #1634 if needed.


thelindat:
iirc source is empty string when triggering disarm from the server. Though I guess if other scripts trigger the disarm event then weapon isn't removed from the inventory.

PR #1698, opened 2024-05-12
fix(client): check if vehicle is locked in 'Inventory.CanAccessTrunk'

Gamzky:
Currently, when you open a trunk and the owner locks that vehicle, the inventory will stay open and you are able to move items. This PR will change it so the trunk closes when the vehicle gets locked.
Also moved the 'vehicle_locked' notification, because else it wouldn't get shown.

### DIFF
--- a/modules/inventory/client.lua
+++ b/modules/inventory/client.lua
@@ -44,6 +44,10 @@ function Inventory.CanAccessTrunk(entity)
         end
     end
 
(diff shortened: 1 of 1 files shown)


thelindat:
I actually need to move this to a server-side check anyway, though I appreciate it.

Gamzky:
Even better, thanks

PR #1716, opened 2024-06-03
Feat(bridge/qbcore): Add qb-inventory compatability

FjamZoo:
Added the most important compatability exports for now, will probably go through the rest at a later date.

These are very much untested.


thelindat:
> These are very much untested.

Which is why I didn't just add them like this in the first place - the functionality and result for these exports need to match qb-inventory or it's pointless.

dorlevy1:
i suggest you to change instead of exports.ox_inventory:
to Inventory.

little bit strange to export from own resource

dorlevy1:
also in CanCarryItem you should add the metadata 
because they have the ability to change the weight through that if they want to

PR #1722, opened 2024-06-11
 fix(keybind): txAdmin default hotkey support

t0r3tto:
fixes the default txAdmin hotkey being tab with the inventory's, no matter the framework (ox_core same issue)


thelindat:
Yeah, no. I used to disallow opening the inventory when nui already had focused - it resulted in constant complaints.

PR #1732, opened 2024-06-21
fix(client): Prevent weapon disarm if weapon is currently being used

spoty123:
Currently, you can unlimitedly use ammo by unequipping the weapon you are using while shooting with it.
This PR solves this bug by preventing weapons from being unequipped while they are actively being used (aka weapon timer exists, and is bigger than 0).

Here is a demonstration video: https://streamable.com/oeoqjz

This "fix" would still not cover a case where the weapon is actively used and a "disarm" is called from elsewhere, but I guess that's a very rare edge case which should be handled by the calling script.

### DIFF
--- a/client.lua
+++ b/client.lua
@@ -513,6 +513,12 @@ local function useSlot(slot, noAnim)
 			if IsCinematicCamRendering() then SetCinematicModeActive(false) end
 
 			if currentWeapon then
+				local weaponAmmo = currentWeapon.metadata.ammo
+
+				if weaponAmmo then
+					TriggerServerEvent('ox_inventory:updateWeapon', 'ammo', weaponAmmo) -- Update the ammo, incase the weapon is still being used right now.
+				end
+
 				local weaponSlot = currentWeapon.slot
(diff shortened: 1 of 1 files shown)


thelindat:
Cannot reproduce.

spoty123:
> Cannot reproduce

I can still very easily reproduce this with the latest release, I'll write the steps, because maybe my previous explanation wasn't very clear.

1. Give yourself any weapon and two magazines worth of ammo. (e.g WEAPON_CARBINERIFLE, x60 ammo-rifle)
2. Reload the weapon. (you will be left with 30 ammo-rifle, as 30 are inside the weapon).
4. Equip the weapon, and start shooting.
5. While shooting (most important), holster it using the hotbar.
6. Equip the weapon, and reload it.
7. The weapon will be reloaded, but no bullets would be removed from the inventory. (the ammo metadata of the weapon still shows as if the weapon wasn't reloaded).

> and this change could easily cause other issues

What issues do you think it could cause? I'll gladly try and come up with another fix.

PR #1750, opened 2024-07-16
Rotate CustomDrop Model

DomScripts:
The goal of this PR is aimed at adding the ability to rotate the entity spawned when using `CustomDrop()`. This was done by adding a rotation to the entity along with a coordinate offset. The rotation and coordinate offset must be done after `PlaceObjectOnGroundProperly()` as it seems to reset the entity's rotation after calling in testing calling the native after applying the offsets. 

Using the native `SetEntityRotation()` requires a rotationOrder. 5 was chosen as it is the standard XYZ but different rotational orders can be found [here](https://docs.fivem.net/natives/?_0xAFBD61CC738D9EB9) if a different one is prefered.

Handling the case if the user doesn't want to set a rotation and offset was also considered by giving a default vec3(0,0,0) in it's place which would make the entity spawned use it's default values. This was chosen to be done when the point is created instead of the time when the entity is created so the entity's offset and rotation can be stored in the points data if someone would ever need that information.

### DIFF
--- a/client.lua
+++ b/client.lua
@@ -1032,6 +1032,8 @@ local function onEnterDrop(point)
 
 		SetModelAsNoLongerNeeded(model)
 		PlaceObjectOnGroundProperly(entity)
+		SetEntityCoords(entity, point.coords.x + point.offset.x, point.coords.y + point.offset.y, point.coords.z + point.offset.z)
+		SetEntityRotation(entity, point.rot.x, point.rot.y, point.rot.z, 5, true)
 		FreezeEntityPosition(entity, true)
 		SetEntityCollision(entity, false, true)
 
@@ -1054,7 +1056,9 @@ local function createDrop(dropId, data)
 		distance = 16,
 		invId = dropId,
 		instance = data.instance,
-		model = data.model
+		model = data.model,
+		rot = data.rot,
+		offset = data.offset,
 	})
 
 	if point.model or client.dropprops then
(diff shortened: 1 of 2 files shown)


Hakkodevelopment:
Seems useless, ngl

thelindat:
What's the point of an offset when you're already setting the coords of the drop?

DomScripts:
The point of an offset is to properly set whatever model you are using on the ground properly. During testing, if you apply the rotation to the entity then call `PlaceObjectOnGroundProperly()` it will reset the rotation of the entity in order to place it on the ground correctly. Therefore you need to apply the rotation after calling the native. The issue you run into then is that if you rotate the entity after setting the coordinates it will no longer be on the ground properly. 

This image below showcases the behavior if you apply a rotation after calling `PlaceObjectOnGroundProperly()` without an offset. The model is no longer on the ground correctly after apply a -90.0 rotation in the X-axis.
image

This image below shows the behavior if you apply the rotation after calling `PlaceObjectOnGroundProperly()` with an offset to the model. 
image

I'm not familiar with any other native that places the object on the ground correctly while keeping the rotation of the object. Just using the native `PlaceObjectOnGroundProperly()` isn't enough in order to achieve this so my approach was to add a coordinate offset to the model in order for people to be able to lineup the models correctly with the ground.


When it comes to the default rotation and offset values. I have removed them and replaced them with checks if they are defined instead in order to not have the useless objects.

PR #1757, opened 2024-07-31
License Buying

cjosephdc:
Ensure that the TextUi is removed when you move away from the buy license point. Additionally, use the license label in the TextUi prompt for improved grammar.

### DIFF
--- a/client.lua
+++ b/client.lua
@@ -1270,7 +1270,7 @@ RegisterNetEvent('ox_inventory:setPlayerInventory', function(currentDrops, inven
 
 		if point.isClosest and point.currentDistance < 1.2 then
 			if not hasTextUi then
-				hasTextUi = nil
+				hasTextUi = point
 				lib.showTextUI(point.message, uiOptions)
 			end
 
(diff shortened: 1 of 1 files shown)


client.lua
```diff
@@ -1300,7 +1300,7 @@ RegisterNetEvent('ox_inventory:setPlayerInventory', function(currentDrops, inven
 			price = data.price,
 			invId = id,
 			nearby = nearbyLicense,
-			message = ('**%s**  \n%s'):format(locale('purchase_license', data.name), locale('interact_prompt', GetControlInstructionalButton(0, 38, true):sub(3)))
+			message = ('**%s**  \n%s'):format(locale('purchase_license', data.label), locale('interact_prompt', GetControlInstructionalButton(0, 38, true):sub(3)))
```
thelindat, line 1303: ... and if there _is_ no label assigned (as default)? Your label might also be "Weapon License", making the prompt "Weapon License license".

client.lua
```diff
@@ -1270,7 +1270,7 @@ RegisterNetEvent('ox_inventory:setPlayerInventory', function(currentDrops, inven
 
 		if point.isClosest and point.currentDistance < 1.2 then
 			if not hasTextUi then
-				hasTextUi = nil
+				hasTextUi = point
```
thelindat, line 1273: PR's should target a single issue.

client.lua
cjosephdc: Oh, I didn't realize that. I've been editing this for a long time with every update. I forgot that I just added a label to my license table.

PR #1763, opened 2024-08-06
feat: qbx_core support

solareon:
Adds support for qbx_core. Blocked on https://github.com/Qbox-project/qbx_core/pull/528


solareon:
Good to merge now. Qbx core has updates completed

PR #1771, opened 2024-08-17
fix(bridge/qbx): sync inventory items to player object

solareon:
I guess some people look at the items table inside the PlayerData.


solareon:
Reworking this

PR #1774, opened 2024-08-18
fix(bridge/qbx): default NPC vehicle database id to nil

Manason:
This fixes an issue where the plate prefix would resolve via a partial match to the database id in the SQL query. This resulted in random NPCs vehicles providing illegal inventory access to player owned vehicles based off a partial plate match to the id.

### DIFF
--- a/modules/bridge/qbx/server.lua
+++ b/modules/bridge/qbx/server.lua
@@ -101,5 +101,5 @@ end
 ---@return number | string
 ---@diagnostic disable-next-line: duplicate-set-field
 function server.getOwnedVehicleId(entityId)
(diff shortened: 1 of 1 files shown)


BerkieBb:
The line I sent under this will error out once you implement this and that results in vehicle trunks and gloveboxes not opening
https://github.com/overextended/ox_inventory/blob/main/modules/inventory/server.lua#L812

mafewtm:
Tested this on `2.42.0` with the convar properly set to `"qbx"`. Both vehicles spawned using `/car` and npc vehicle trunks and gloveboxes open without issue

Issue #1699, opened 2024-05-12
Notice: Limited feature/enhancements updates


thelindat:
No major enhancements or features will be implemented or merged from requests.

Ideally any bugs or minor nuisances can be patched, or other small changes can be implemented infrequently - the idea is to be able to provide long-term support and ensuring a stable release is always available.

I want to be able to work on ox_inventory v3 at some point which will guarantee many breaking changes and will require a great deal of effort to help users transition towards it.

---

**Some expected changes for v3**
- Improved third-party support, with externally registered items, crafting recipes, etc.
- Improved separation of data and code (no functions in data files); using json and the database
- Improved database structure
  - register at least some generic item, stash, and shop data in the database
  - store inventory slots per-row to improve indexing and saving, especially reducing the amount of needless data being sent to the database

CeebDev:
Some suggestions for the v3

- Be able to use / interact with items directly in containers 
	Ex: 
	- Cigarette pack = container with 20 cigarettes, be able to use one of them instead of swipe it to the main inv)
	- Wallet = container with money, be able to pay in shops etc with the money inside wallet
- Select payment methods in shops
- Check degrade for ingredients in craft benchs

thelindat:
> Be able to use / interact with items directly in containers

This would definitely need to be a case-by-case basis. 

> Wallet = container with money, be able to pay in shops etc with the money inside wallet

Containers definitely need a rework and I need an improve "search" function that takes some different options, including going through containers. Would get a bit shitty if we have nested containers, however.

> Steal someone = be able to search inside containers of robbed player and take items

Pretty much comes down to needing to return to the previous inventory, or otherwise making containers open a new inventory window.

> Select payment methods in shops

Already something I've wanted to do for a long time.

> Check degrade for ingredients in craft benchs

Crafting system needs a complete overhaul.

> Add containers with predefined items in shop (Ex: Cigarette pack)

I feel like hooks are good enough for this; also need to consider - does this need a full inventory or just some metadata (x cigarettes remaining). On the matter of hooks, the new inventory will be built around them to allow far more actions; a lot of hardcoded behaviour will moved to hooks.

TonybynMp4:
> > Be able to use / interact with items directly in containers
> 
> This would definitely need to be a case-by-case basis.

Could be done by adding something similar to stack & close in data/items.lua (or whatever it becomes) 🤔  

Also *would* it be possible to have tabs for the inventories?
[...]

PR #1782, opened 2024-08-24
Add the option to lock a stash to a certian item

### LINKED ISSUE #1781: Add the option to lock stashes to an item
**Is your feature request related to a problem? Please describe.**
Issue is not related to a problem but is soley our servers need, so I thought why not share it.

**Describe the solution you'd like**
So currently you can lock a stash to a group (job) and that isn't practical if multiple people share for example a house. So adding a canInteract and checking if the player has the item we want is the sollution I propose.

**Describe alternatives you've considered**
The althernative is to add all the jobs the players have but that isn't really practical.

**Additional context**

Mytic2330:
If specified in stashes.lua, lock a stash to a certian item or list of items.
Using the canInteract function from ox_target to check if player has the certian item/s in the inventory.

Only implemented the check on client side and can only be set before the script start and not at runtime.

### DIFF
--- a/modules/inventory/client.lua
+++ b/modules/inventory/client.lua
@@ -350,6 +350,22 @@ Inventory.Stashes = setmetatable(lib.load('data.stashes'), {
                                 icon = stash.target.icon or 'fas fa-warehouse',
                                 label = stash.target.label or locale('open_stash'),
                                 groups = stash.groups,
+								canInteract = function()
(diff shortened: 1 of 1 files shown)


thelindat:
Something with no server security is pointless.
Regardless - https://github.com/overextended/ox_inventory/issues/1699

PR #1802, opened 2024-09-13
Item amount bug after crafting

hearondale:
- Synchronization of slots, that despite being used in craft still have same amount in UI as before the craft (the amount is not being updated for the UI, unless the slot with item is dragged). 
For me, It didn't work when and not the whole item, but only a part of its durability was used and therefore a new item with new durability was added to another slot, but the amount of items in the initial one remained the same as before craft.

### DIFF
--- a/modules/crafting/server.lua
+++ b/modules/crafting/server.lua
@@ -213,7 +213,8 @@ lib.callback.register('ox_inventory:craftItem', function(source, id, index, reci
 							end
 
 							invSlot.count -= 1
-                            invSlot.weight -= item.weight
+                            invSlot.weight = Inventory.SlotWeight(item, invSlot)
+
 							left:syncSlotsWithClients({
 								{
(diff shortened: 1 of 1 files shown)


modules/crafting/server.lua
```diff
@@ -213,6 +213,16 @@ lib.callback.register('ox_inventory:craftItem', function(source, id, index, reci
 							end
 
 							invSlot.count -= 1
+							left:syncSlotsWithClients({
+								{
+									item = invSlot,
+									inventory = left.id
+								},
+								{
+									item = invSlot,
```
thelindat, line 222: twice?

thelindat: Actually need to update the slot weight after setting the count, too.

hearondale: wasn't able to find where the weight gets updated

PR #1830, opened 2024-11-04
RegisterCraftingBench

lokkafr:
Created the `RegisterCraftingBench` export server-side so that developers can create benches dynamically in the same was as stashes. I didn't concern myself with security for the client event as the main scripts code handles distance checks and checking if a bench exists.

### DIFF
--- a/modules/crafting/client.lua
+++ b/modules/crafting/client.lua
@@ -96,7 +96,6 @@ end
 
 RegisterNetEvent('ox_inventory:registerCraftingBench', function(id, data)
     createCraftingBench(id, data)
-	return true
 end)
 
 for id, data in pairs(lib.callback.await('ox_inventory:getCraftingBenches', false)) do createCraftingBench(id, data) end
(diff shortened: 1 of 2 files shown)


lokkafr:
Through testing, I just found an error where, due to the await, the inventory does not ready in time on the client side. Simply corrected by changing the callback.

modules/crafting/client.lua
```diff
@@ -94,6 +94,10 @@ local function createCraftingBench(id, data)
 	end
 end
 
-for id, data in pairs(lib.load('data.crafting')) do createCraftingBench(id, data) end
+RegisterNetEvent('ox_inventory:registerCraftingBench', function(id, data)
+    createCraftingBench(id, data)
+end)
+
+lib.callback('ox_inventory:getCraftingBenches', false, function(data) for id, data in pairs(data) do createCraftingBench(id, data) end end)
```
iampitchou, line 101: I would rename the var `data` in the loop
```suggestion
lib.callback('ox_inventory:getCraftingBenches', false, function(data) for id, bench in pairs(data) do createCraftingBench(id, bench) end end)
```

PR #1831, opened 2024-11-04
add useItem Hook

BaziForYou:
added useItem Hook for some reason I found it missing like making logs or making other scripts than the original script made usable can add more checks on that and after that, it will make more potential on creating new things and if this gets merged I will fix doc page to add this part


thelindat:
https://github.com/overextended/ox_inventory/commit/70abcfd3ab7d4e6d3dd6e852712cc0467e864ab6

PR #1841, opened 2024-11-21
fix(server/mysql): fix stashes saving string

Frowmza:
previous state would throw `bad argument #1 to 'format' (number expected, got string)` as it expect the first argument to be number and not string

### DIFF
--- a/modules/mysql/server.lua
+++ b/modules/mysql/server.lua
@@ -268,7 +268,7 @@ function db.saveInventories(players, trunks, gloveboxes, stashes, total)
                         end
                     end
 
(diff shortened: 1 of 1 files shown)


thelindat:
Swear this was corrected forever ago 🤔

fwiw nobody should be disabling bulk stash saves.

Frowmza:
I discovered this while working with a server. We can expect anything when doing commissions, so I’m wondering why they’re disabling it. Also, could I get a review on the ox_lib PR about the Timer class when you free? I really need it for a resource. thank you

PR #1856, opened 2024-12-20
fix(server): update entity and plate check logic

RrybaN:
Fixes a script error caused by nil being passed to `string.match` when `DoesEntityExist(invEntity)` returns false.


CeebDev:
When `GetVehicleNumberPlateText` can return nil ?

RrybaN:
I’ve updated my PR. The issue was that even though `DoesEntityExist(invEntity)` returned false, `GetVehicleNumberPlateText(invEntity)` was still called, which returned `nil` and caused the error.

image

PR #1863, opened 2025-01-02
esx: 1.6.0 to 1.12

userMacieG:


Mycroft-Studios:
not needed. since the minimum is still 1.6

PR #1875, opened 2025-01-26
Support for entity lockdown with dumpsters

Scullyy:
This implementation has been tested extensively and worked flawlessly during all tests. The only limitation is that dumpsters need to be frozen when nearby to maintain a consistent position across clients. Given that dumpsters in the game tend to move like cardboard boxes, this could also be considered an improvement haha.

If anyone has suggestions to improve this further, feedback is welcome!

### DIFF
--- a/server.lua
+++ b/server.lua
@@ -67,6 +67,25 @@ end
 exports('setPlayerInventory', server.setPlayerInventory)
 AddEventHandler('ox_inventory:setPlayerInventory', server.setPlayerInventory)
 
+local registeredDumpsters = {}
+
+---@param coords vector3
+---@return string?
+local function getDumpsterFromCoords(coords)
+	local found
+
+	for i = 1, #registeredDumpsters do
+		local distance = #(coords - registeredDumpsters[i])
+
+		if distance < 0.1 then
+			found = i
+			break
+		end
+	end
+
+	return found
+end
+
 ---@param playerPed number
 ---@param stash OxInventory
 ---@return vector3?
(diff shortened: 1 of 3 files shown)


Scullyy:
Showcase of it working: https://streamable.com/crqll0

server.lua
```diff
@@ -67,6 +67,25 @@ end
...
+
+		if distance < 0.1 then
+			found = i
+			break
+		end
+	end
+
+	return found
+end
+
```
thelindat, line 88: If you're freezing dumpsters then you might as well index them by the coords. No need to find the nearest.

server.lua
```diff
@@ -178,17 +197,13 @@ local function openInventory(source, invType, data, ignoreSecurityChecks)
...
-			---@cast data string
-			right = Inventory(data)
+			local dumpsterId = getDumpsterFromCoords(data)
+			right = dumpsterId and Inventory(('dumpster-%s'):format(dumpsterId))
 
 			if not right then
-				local netid = tonumber(data:sub(9))
-
-				-- dumpsters do not work with entity lockdown. need to rewrite, but having to do
-				-- distance checks to some ~7000 dumpsters and freeze the entities isn't ideal
```
thelindat, line 188: Just going to point at

> freeze the entities isn't ideal

thelindat (changes requested):
This should be tied behind a convar rather than replacing the existing system. Neither is ideal anyway.

server.lua
Scullyy: There has been some weird cases where the coords were off by a tiny bit while testing with multiple clients, I originally had it how you suggested but had to implement the function to resolve the issue.

server.lua
Scullyy: The distance check to ~7000 dumpsters has been avoided by only doing it to those that have been interacted with, unfortunately there isn't really a way to get around this without freezing them but I will implement the convar as suggested.

Scullyy:
If you would like me to rename the convar just let me know, I wasn't really sure what to name it but considered this "fitting".

thelindat:
> If you would like me to rename the convar just let me know, I wasn't really sure what to name it but considered this "fitting".

Probably something like `inventory:networkdumpsters`.

PR #1890, opened 2025-02-12
Tweak: Use locales for weight tag

Maximus7474:
Noticed that the `kg` and `g` where hard coded, added them to locales files and made sure they were called properly.


Maximus7474:
Will test again in the morning, but I'm pretty sure I didn't miss anything.

thelindat:
Just swapping out kg/g for lbs/oz makes no sense, and well implementing this isn't worthwhile.

PR #1894, opened 2025-02-23
Added pre-filling item containers

Zackaery:
This pull request introduces functionality to prepopulate containers when they are created, allowing you to define preset containers with specific items.

This is useful for scenarios such as:

Lootable Crates – Ensure predefined loot spawns in chests or boxes.
Starter Kits – Automatically provide a set of items inside of a box.
Item Packages - Items that would come prepackaged in bulk like a pack of cigarettes that contains 10 cigarettes.
Mission Rewards – Quest-related items directly into an item container.

If a dev passes the items property table in when they define a new item container, it will add those items in the container when the item is received via the AddItem function.

An example was added to the pizzabox container, where it will add 1x Pizza.

### DIFF
--- a/modules/inventory/server.lua
+++ b/modules/inventory/server.lua
@@ -397,6 +397,17 @@ function Inventory.SetSlot(inv, item, count, metadata, slot)
 	inv.items[slot] = currentSlot
 	inv.changed = true
 
+	if metadata.container and metadata.items then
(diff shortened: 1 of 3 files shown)


thelindat:
This is such a weird hacky way of "prefilling" a container. Just add the container item, then get the inventory from the container and add items to that.

PR #1898, opened 2025-02-28
fix(init.lua): drop weight

iSentrie:
Invalid convar for dropweight was set.

# Maximum drop capacity, in grams
setr inventory:dropweight 30000

https://overextended.dev/ox_inventory#config


tom-osborne (approved):
LGTM

PR #1907, opened 2025-03-24
fix(server/inventory): CanCarryItem returning false where an item should be able stack.

### LINKED ISSUE #1906: CanCarryItem doesn't seem to properly handle when metadata is nil
**Describe the bug**
If you have an inventory that is full of stuff (stackables and non-stackables), and then you try and use `CanCarryItem` with a stackable item that's in the inventory and nil metadata, it'll say that you can't carry the item.

Tracked it down to being related to how `{ type = metadata or nil }` table gets treated by GetItemSlots as a hashmap by `table.type`, which is used in the strict metadata check.

**Framework**
QBox

**Resource version**
2.44.1

**To Reproduce**
1. Fill up your inventory except one slot with a non-stackable item.
2. In that one slot, add an item that is stackable *with no metadata*.
3. In `/runcode` or a resource, run the following code, making sure its ran server-sided.
```lua
local inventory = 1
local itemName = "burger"
local count = 10
local metadata = nil
if exports.ox_inventory:CanCarryItem(inventory, itemName, count, metadata) then
    local addItemSuccess, reason = exports.ox_inventory:AddItem(inventory, itemName, count, metadata)
    if addItemSuccess then
        print("added item successfully")
    else
        print("failed to add item", reason)
    end
else
    print("failed to add item, cannot carry item")
end
```

**Expected behavior**
The item that we're trying to add, should've added into the stack in the inventory.

**Screenshots**
Image
*Image shows a full inventory, with the burger being the target item that i was trying to check with `CanCarryItem` The Facade Laptops are non-stackable*

**Additional context**
N/A

gamenew09:
Fixes #1906

----

Essentially when an inventory is full of items, when you try and check if you can add a stackable item that is in the inventory using `CanCarryItem`, it returns false. It seemed to stem from the strict metadata check.

When the metadata value is nil in CanCarryItem, the metadata table created is `{ type = nil }`. Lua treats that new table as a hashmap even though it's really just an empty table (at least `table.type` looks at the table like that), and it seems to break the strict metadata check.

### DIFF
--- a/modules/inventory/server.lua
+++ b/modules/inventory/server.lua
@@ -1394,7 +1394,15 @@ function Inventory.CanCarryItem(inv, item, count, metadata)
 		inv = Inventory(inv) --[[@as OxInventory]]
 
 		if inv then
-			local itemSlots, _, emptySlots = Inventory.GetItemSlots(inv, item, type(metadata) == 'table' and metadata or { type = metadata or nil })
+			local md
+			if type(metadata) == 'table' then
+				md = metadata
+			elseif type(metadata) == 'string' then
+				md = { type = metadata }
+			else
+				md = {}
+			end
+			local itemSlots, _, emptySlots = Inventory.GetItemSlots(inv, item, md)
 
 			if not itemSlots then return end
 


gamenew09:
Hopefully I explained it well, my explaining skills are not the best haha

modules/inventory/server.lua
```diff
@@ -1394,7 +1394,15 @@ function Inventory.CanCarryItem(inv, item, count, metadata)
 		inv = Inventory(inv) --[[@as OxInventory]]
 
 		if inv then
-			local itemSlots, _, emptySlots = Inventory.GetItemSlots(inv, item, type(metadata) == 'table' and metadata or { type = metadata or nil })
+			local md
+			if type(metadata) == 'table' then
+				md = metadata
+			elseif type(metadata) == 'string' then
```
jag3dagster, line 1400: [The `type` value of metadata can be of `any` type](https://overextended.dev/ox_inventory/Guides/metadata#special-metadata-properties), it doesn't have to be a string. I believe checking if metadata is not nil here is the better check

PR #1925, opened 2026-04-28
Added possibility to give items through car windows

tugamars:
In the current iteration of inventory, giving items through windows of vehicles is not possible: i.e. if player A is near the driver window and player B is sitting in driver seat.

This PR adds that (along with https://github.com/overextended/ox_lib/pull/757 ) as a feature.

### DIFF
--- a/client.lua
+++ b/client.lua
@@ -1689,10 +1689,10 @@ local function isGiveTargetValid(ped, coords)
 
     local entity = Utils.Raycast(1|2|4|8|16, coords + vec3(0, 0, 0.5), 0.2)
 
-    if(GetVehiclePedIsIn(ped, false) and not cache.vehicle) then
-		local currCoords=GetWorldPositionOfEntityBone(ped,0);
-		if( #( pedCoords.xy - currCoords.xy ) <= 2.0) then
-			return true;
+    if GetVehiclePedIsIn(ped, false) and not cache.vehicle then
+		local currCoords=GetWorldPositionOfEntityBone(ped,0)
+		if #(pedCoords.xy - currCoords.xy) <= 2.0 then
+			return true
 		end
 	end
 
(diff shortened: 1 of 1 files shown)


client.lua
```diff
@@ -1698,7 +1705,13 @@ RegisterNUICallback('giveItem', function(data, cb)
     if usingItem then return end
 
 	if client.giveplayerlist then
-		local nearbyPlayers = lib.getNearbyPlayers(GetEntityCoords(playerPed), 3.0)
+		local coords=GetEntityCoords(playerPed)
+		
+		if cache.vehicle then
+			coords=GetWorldPositionOfEntityBone(playerPed, 0)
+		end
```
thelindat:
```lua
local coords = cache.vehicle and GetWorldPositionOfEntityBone(playerPed, 0) or GetEntityCoords(playerPed)
```

iSentrie:
Check if window is rolled down maybe?

client.lua
```diff
@@ -1689,6 +1689,13 @@ local function isGiveTargetValid(ped, coords)
 
     local entity = Utils.Raycast(1|2|4|8|16, coords + vec3(0, 0, 0.5), 0.2)
 
+    if GetVehiclePedIsIn(ped, false) and not cache.vehicle then
+		local currCoords=GetWorldPositionOfEntityBone(ped, 0)
+		if #(pedCoords.xy - currCoords.xy) <= 2.0 then
+			return true
+		end
+	end
+
```
thelindat: This is unnecessary, however the server has a distance check that might need to be double-checked to be a bit more lenient.

The raycast above should also be adjusted not to check for vehicles (remove 2 from the flags).

tugamars: Fixed that. The server check has way more wiggle room (15m), and it worked with all the vehicles I tested that are long

PR #1927, opened 2026-04-29
feat(ui): Add thousands separator formatting to item amount input

Acc-Off:
## Summary

Replace the `type="number"` input with a `type="text"` input for the item
amount field in `InventoryControl`, and apply comma-separated thousands
formatting (e.g. `1,000`, `10,000`) using `toLocaleString('en-us')`.

## Changes

- Changed input type from `number` to `text`
- Format the entered value with thousands separators on each keystroke
- Preserve cursor position after formatting using `useRef` and `useEffect`
- Strip non-numeric characters before dispatching the value to the Redux store

## Motivation

The default `number` input does not display thousand separators, making it
difficult to read large quantities at a glance. This change improves
readability without affecting the underlying numeric value stored in state.

## Testing

- Type a large number (e.g. `1000000`) and verify it displays as `1,000,000`
- Verify the cursor does not jump to the end while editing mid-value
- Verify that the dispatched amount is correct (no commas included)
- Verify that empty input resets the amount to `0`

## Demo
563879358-03f1a7ad-6307-42d4-9742-64dd8cd9633d

### DIFF
--- a/web/src/components/inventory/InventoryControl.tsx
+++ b/web/src/components/inventory/InventoryControl.tsx
@@ -9,14 +9,18 @@ import { fetchNui } from '../../utils/fetchNui';
 import { Locale } from '../../store/locale';
 import UsefulControls from './UsefulControls';
 
+const formatAmount = (n: number) => (n ? n.toLocaleString('en-US') : '');
+const digitsOnly = (s: string) => s.replace(/\D/g, '');
+const countDigitsBefore = (s: string, index: number) => digitsOnly(s.substring(0, index)).length;
+
 const InventoryControl: React.FC = () => {
   const itemAmount = useAppSelector(selectItemAmount);
   const dispatch = useAppDispatch();
 
   const [infoVisible, setInfoVisible] = useState(false);
-  const [value, setValue] = useState("");
+  const [value, setValue] = useState(formatAmount(itemAmount));
   const inputRef = useRef<HTMLInputElement>(null);
-  const cursorRef = useRef<{ digitsBefore: number; formattedValue: string } | null>(null);
+  const cursorRef = useRef<number | null>(null);
 
   const [, use] = useDrop<DragSource, void, any>(() => ({
     accept: 'SLOT',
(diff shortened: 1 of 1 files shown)


Maximus7474:
There's an issue with the implementation, when pressing backspace or delete around a comma it resets the cursor to the end of line.

issue

Acc-Off:
Thank you for the feedback! I've pushed a fix that addresses this.

Previously, pressing Backspace or Delete adjacent to a comma only moved the cursor without modifying the value, causing it to reset to the end. The updated behavior now deletes the comma together with the adjacent digit:

Backspace after a comma (e.g. "12,|345") → removes 2, → "1,345"
Delete before a comma (e.g. "12|,345") → removes ,3 → "1,245"

Acc-Off:
@Maximus7474 
Not sure if GitHub notified you automatically, so mentioning you just in case! 
I've pushed a fix for the backspace/delete issue you mentioned. Thanks!

unitysync (changes requested):
this can be done with significantly less code. something like this would be good. (ive truncated the unchanged code, however it should be a simply copy and paste into the file)

```tsx
const formatAmount = (n: number) => (n ? n.toLocaleString('en-US') : '')
const digitsOnly = (s: string) => s.replace(/\D/g, '')
const countDigitsBefore = (s: string, index: number) => digitsOnly(s.substring(0, index)).length

const itemAmount = useAppSelector(selectItemAmount)
const dispatch = useAppDispatch()

const [infoVisible, setInfoVisible] = useState(false)
const [value, setValue] = useState(formatAmount(itemAmount))

const inputRef = useRef<HTMLInputElement>(null)
const cursorRef = useRef<number | null>(null)

const commitValue = (raw: string, cursorIndex: number) => {
  const digitsBefore = countDigitsBefore(raw, cursorIndex)
  const num = parseInt(digitsOnly(raw), 10) || 0

  setValue(formatAmount(num))
  dispatch(setItemAmount(num))
  cursorRef.current = digitsBefore
}

// use `handleChange` instead of `inputHandler` for naming consistency
const handleChange = (event: React.ChangeEvent<HTMLInputElement>) =>
  commitValue(event.target.value, event.target.selectionStart ?? 0)

const handleKeyDown = (event: React.KeyboardEvent<HTMLInputElement>) => {
  const el = event.currentTarget
  const pos = el.selectionStart ?? 0

  if (pos !== el.selectionEnd) return

  if (event.key === 'Backspace' && el.value[pos - 1] === ',') {
    event.preventDefault()
    commitValue(el.value.slice(0, pos - 2) + el.value.slice(pos), pos - 2)
  } else if (event.key === 'Delete' && el.value[pos] === ',') {
    event.preventDefault()
    commitValue(el.value.slice(0, pos) + el.value.slice(pos + 2), pos)
  }
}

useEffect(() => {
  if (!inputRef.current || cursorRef.current === null) return
  let newPos = 0
  let count = 0

  for (let i = 0; i < value.length && count < cursorRef.current; i++) {
    if (/\d/.test(value[i])) count++
    newPos++
  }

  inputRef.current.setSelectionRange(newPos, newPos)
  cursorRef.current = null
}, [value])```

**Comment by Acc-Off** [comment]
Thanks for the clean refactor suggestion! Applied it — much simpler now with commitValue centralizing the logic. Pushed!

File: web/src/components/inventory/InventoryControl.tsx [hunk 1]
```diff
@@ -9,14 +9,18 @@ import { fetchNui } from '../../utils/fetchNui';
 import { Locale } from '../../store/locale';
 import UsefulControls from './UsefulControls';
 
+const formatAmount = (n: number) => (n ? n.toLocaleString('en-US') : '');
```
unitysync: i previously overlooked this. to avoid displaying a blank box, the default value should be "0".
```ts
const formatAmount = (n: number) => (n > 0 ? n.toLocaleString('en-US') : '0');
```

unitysync (changes requested):
looks good so far, except i noticed something in testing.

web/src/components/inventory/InventoryControl.tsx
Acc-Off: Fixed — formatAmount now returns '0' for falsy values to match the original behavior. Pushed!

unitysync (approved):
LGTM. @thelindat

FjamZoo (approved):

PR #1932, opened 2026-05-06
fix(server/crafting): validate ingredient slots have not been replaced during craft progress

Kenshiin13:
## Summary

`craftItem` had a distinct bug that results in a free-craft exploit:

1. **Slot identity was dropped between build and consume.** The ingredient slot list (`tbl`) is built before awaiting the client's progress confirmation, then consumed from afterward. `tbl` keyed by slot id only, with the count as the sole payload — no name, no durability snapshot. Post-await consumption trusted whatever currently occupied the slot. The await spans `recipe.duration or 3000` ms, during which `ox_inventory:swapItems` (a plain server callback) can replace the slot's contents.

## The wrong code

`modules/crafting/server.lua:167-189` — `tbl` was keyed by slot id with the count as the only payload:

```lua
tbl[slot.slot] = needs
-- ...
tbl[slot.slot] = slot.count
```

`modules/crafting/server.lua:201` — `lib.callback.await` yields for `recipe.duration or 3000` ms.

`modules/crafting/server.lua:208-251` — post-await consumption used whatever currently occupied the slot:

```lua
for slot, count in pairs(tbl) do
    local invSlot = left.items[slot]
    if not invSlot then return end

    if count < 1 then
        local item = Items(invSlot.name)
        local durability = invSlot.metadata.durability or 100
        -- ...consumes whatever item / durability is in the slot now
    else
        local removed = invSlot and Inventory.RemoveItem(left, invSlot.name, count, nil, slot)
        if not removed then return end
    end
end
```

`Inventory.RemoveItem` with a `slot` argument also doesn't validate that the slot has at least `count`; it calls `SetSlot(-count)` on whatever's there and reports `removed = count` regardless, so a slot with fewer items than required still yields a "successful" craft.

## Holes the trust pattern opens

- **Different item entirely.** Slot now holds something else; `RemoveItem(invSlot.name, ...)` consumes the substitute and the recipe still produces its output.
- **Same name, fewer items.** Slot has e.g. 2 instead of 5; `SetSlot(-5)` clears the 2 and reports success for 5.
- **Fractional consume, lower remaining durability.** The build-time eligibility check (`durability >= needs * 100`) is not re-evaluated; an item below the threshold gets consumed.
- **Fractional consume, item without durability.** `or 100` defaults a missing durability to 100, bypassing the original eligibility.
- **Fractional consume, time-based degrade with insufficient remaining percentage.** The percentage recompute from build time isn't repeated post-await.
- **Hash-order ingredient skip.** A `needs == 0` sentinel iterating before consumed ingredients ends the entire ingredient loop; consumed ingredients past that point are never reserved or removed.

## Fix

- Store the expected ingredient `{ name, count }` in `tbl` instead of just the count.
- Extract the durability/percentage eligibility into a helper used at both build and consume, so the math has one source of truth.
- Re-run that eligibility check against the current slot post-await; abort if it doesn't pass.
- Add a count check before slot-bound `RemoveItem` so a smaller stack can't masquerade as a successful removal.
- `Utils.LogExploit` on the name-mismatch path (the unambiguously-malicious one). Other abort paths stay silent because they could in principle stem from cross-resource state mutation. Drop flag omitted (= `false`); flip to `true` if you want auto-kick.

The redundant pre-loop existence check (`for name, needs in pairs(recipe.ingredients) do if Inventory.GetItemCount(left, name) < needs then return end end`) was removed; the per-slot loop is strictly stronger now and the total-count scan is dead weight.

## Behavior matrix

| Slot state at post-await | Pre-fix | Post-fix |
| --- | --- | --- |
| Unchanged (honest craft) | works | works |
| Slot empty | aborts | aborts |
| Slot now holds different item | consumes substitute, craft succeeds | aborts (logs as exploit) |
| Same name, fewer items than required | partial removal, craft succeeds | aborts |
| Fractional, lower remaining durability | consumes anyway | aborts |
| Fractional, item with no durability | defaults to 100, consumes | aborts |
| Fractional, time-based degrade insufficient % | consumes anyway | aborts |
| Recipe with `needs == 0` sentinel iterated before consumed ingredients | consumed ingredients silently skipped, free craft | sentinel skipped, consumed ingredients still reserved |

## Out of scope for this PR (worth follow-ups):
- `Inventory.RemoveItem` with a slot argument should validate `slotData.count >= count`; the band-aid is in the call site here.
- `craftItem` could lock ingredient slots in `activeSlots` for the duration of the await (matching `swapItems`' pattern), making post-await re-validation unnecessary.

### DIFF
--- a/modules/crafting/server.lua
+++ b/modules/crafting/server.lua
@@ -164,7 +164,7 @@ lib.callback.register('ox_inventory:craftItem', function(source, id, index, reci
 			table.wipe(tbl)
 
 			for name, needs in pairs(recipe.ingredients) do
-				if needs == 0 then goto continue end
+				if needs == 0 then break end
 
 				local slots = items[name] or items
 
@@ -194,8 +194,6 @@ lib.callback.register('ox_inventory:craftItem', function(source, id, index, reci
 					-- Player does not have enough items (ui should prevent crafting if lacking items, so this shouldn't trigger)
 					if needs > 0 and i == #slots then return end
 				end
-
-				::continue::
 			end
 
 			if not TriggerEventHooks('craftItem', {
--- a/modules/crafting/server.lua
+++ b/modules/crafting/server.lua
@@ -164,7 +164,7 @@ lib.callback.register('ox_inventory:craftItem', function(source, id, index, reci
 			table.wipe(tbl)
 
 			for name, needs in pairs(recipe.ingredients) do
-				if needs == 0 then break end
+				if needs == 0 then goto continue end
 
 				local slots = items[name] or items
 
@@ -194,6 +194,8 @@ lib.callback.register('ox_inventory:craftItem', function(source, id, index, reci
 					-- Player does not have enough items (ui should prevent crafting if lacking items, so this shouldn't trigger)
 					if needs > 0 and i == #slots then return end
 				end
+
+				::continue::
 			end
 
 			if not TriggerEventHooks('craftItem', {
--- a/modules/crafting/server.lua
+++ b/modules/crafting/server.lua
@@ -167,19 +167,19 @@ lib.callback.register('ox_inventory:craftItem', function(source, id, index, reci
 								local percentage = ((durability - os.time()) * 100) / degrade
 
 								if percentage >= needs * 100 then
-									tbl[slot.slot] = needs
+									tbl[slot.slot] = { name = name, count = needs }
 									break
 								end
 							else
-								tbl[slot.slot] = needs
+								tbl[slot.slot] = { name = name, count = needs }
 								break
 							end
 						end
 					elseif needs <= slot.count then
-						tbl[slot.slot] = needs
+						tbl[slot.slot] = { name = name, count = needs }
 						break
 					else
-						tbl[slot.slot] = slot.count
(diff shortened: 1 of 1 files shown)


Maximus7474 (changes requested):
If possible, let's add the [`Utils.LogExploit`](https://github.com/overextended/ox_inventory/blob/main/modules/utils/server.lua#L70-L80) before escaping the craft if the item name does not align, not sure if having the drop flag as false is appropriate.

thelindat:
Unless somebody changed something, inventory should have been inaccessible by anyone during the crafting process. Are you manually triggering `ox_inventory:swapItems`?

Kenshiin13:
> Are you manually triggering `ox_inventory:swapItems`?

Yes

thelindat:
I've corrected some behaviour to ensure the player inventory is always properly updated to closed on the server.
swapSlots will now reject any swap actions when the player inventory is closed.

I've added a new lock manager which takes any number of ids and generates a lock if none of the ids are being used by an existing lock. It acts essentially the same as the activeSlots table used by swapSlots and giveItem - both of which now use the lock manager. The same locks are now used during crafting.

PR #1284, opened 2023-04-27
fix(qb/client): prevent inventory access while in last stand state

Manason:
fixes #1283 

Untested

Issue #65, opened 2021-05-13
Inventory with esx_policejob searching and put to the stocks


DexikCZ123:
Not working search people inventory a don work give item to the stocks

Ruqen:
What do you mean? I can't understand you

thelindat:
You're going to have a bad time trying to use default inventory menus with this.

Go to documentation for replacing the default police body search.  
Disable the base police inventory stuff and use the stashes. There are personal lockers for players, and /evidence # while near the evidence room (set the positions in `shared/stashes.lua`)

PR #1935, opened 2026-05-11
feat(bridge/qb): restore qb-core bridge

JensGryspeert:
## Why

`modules/bridge/qb` was removed in [34a3e9b](https://github.com/overextended/ox_inventory/commit/34a3e9b) ("fix: cure cancer", between v2.40.2 and v2.41.0). Stock `qb-core` servers (not `qbx_core`) lost upstream support. This PR brings the bridge back so `setr inventory:framework qb` boots ox_inventory against `qb-core` again.

Opened as a draft per CONTRIBUTING.md — happy to close if a re-add isn't wanted.

## What's restored

- **`modules/bridge/qb/{client,server}.lua`** — restored verbatim from `34a3e9b~1`.
- **`modules/mysql/server.lua`** — qb branch (`players.citizenid`, `player_vehicles.plate`).
- **`modules/items/server.lua`** — `QBCore.Shared.Items` auto-import into `data/items.lua`.
- **`setup/convert.lua`** — `ConvertQB` migration (player inventory/money, trunks, gloveboxes, stash items, evidence drawers).
- **`server.lua`** — `qb` re-added to the `convertinventory` valid-args list and `UseItem` comment.
- **`modules/pefcl/server.lua`** — qb cash/money mismatch note.

All qb bridge bugfixes that landed before the deletion are preserved: #1284, #1285, #1411, #1587, #1608, #1711.

## Bugs fixed in this PR

1. **`onResourceStart` never stopped `qb-weapons` / `qb-shops`** — pre-existing logic used `or` instead of `and`, so the guard short-circuited and the resources kept running. Fixed in the restored handler.
2. **`Inventory.RemoveItem` errored when a slot was fully emptied** — `Inventory.SetSlot` returns `nil` after clearing a slot (e.g. removing the last `pressurewash` in slot 1). The error-checking added in [067e89f8](https://github.com/overextended/ox_inventory/commit/067e89f8) treated that `nil` as failure and threw `Failed to remove 1x <item> from inventory-<id>:slot-<n> (nil).` Fix: `SetSlot` now returns `true` when it intentionally empties a slot, so the success path stays truthy.

## Contract check

The restored bridge still matches the current ox_inventory surface:

| Surface | Status |
|---|---|
| `server.hasGroup` numeric-rank specs (#65) | qb writes via `grade.level` — compatible |
| `server.syncInventory(inv)` | shape unchanged, reads `inv.id` |
| `server.setPlayerData` return shape | additive (`job` / `gang`) |
| `Inventory.{GetItem,AddItem,RemoveItem,SetItem,GetSlot*,Clear,Save,Search,CanCarryAmount,SlotWeight,GetAccountItemCounts}` | signatures match |
| `Items.Metadata` | signature matches |
| `lib.load('modules.bridge.qb.server')` | works unmodified |

## How to test

- [ ] Boot FiveM server with `setr inventory:framework qb` and latest `qb-core` — no errors at `init.lua`.
- [ ] Connect a character — F2 opens inventory, slots and weight defaults apply.
- [ ] `/setmoney cash 500` → `money` item shows 500. Drop it → cash decreases.
- [ ] `/setjob police 3` → group-restricted shop accessible. Demote to `police 1` → denied.
- [ ] Use a registered item → `QBCore.Functions.CanUseItem` callback fires.
- [ ] Equip a weapon item; `qb-weapons` auto-stops on boot.
- [ ] `/cuff` and last-stand state → inventory blocked (#1284 / #1285 behaviour).
- [ ] License purchase at PD desk → `PlayerData.metadata.licences` updates; repeat returns `already_have`.
- [ ] Remove an item from `data/items.lua`, restart → re-imported from `QBCore.Shared.Items`.
- [ ] `exports.ox_inventory['qb-inventory.HasItem'](src, 'bread', 1)` → `true`.
- [ ] `convertinventory qb` against a qb-core DB snapshot → ox-format JSON for players, trunks, gloveboxes, evidence drawers.
- [ ] Remove the **last** unit of an item from a slot → no `SCRIPT ERROR` (regression check for the SetSlot fix).

## Known race

`setupPlayer` from `PlayerLoaded` can fire before the `SetTimeout(500, …)` that assigns `QBCore`. The same timeout also scans `GetQBPlayers()` so existing players are picked up, but a connection inside that window can hit a nil `QBCore`. Pre-existing behaviour from the original bridge — happy to tighten with a `lib.waitFor`-style guard if preferred.

### DIFF
--- a/modules/bridge/qb/client.lua
+++ b/modules/bridge/qb/client.lua
@@ -1,0 +1,85 @@
+local QBCore = exports['qb-core']:GetCoreObject()
+local Inventory = require 'modules.inventory.client'
+local Weapon = require 'modules.weapon.client'
+
+RegisterNetEvent('QBCore:Client:OnPlayerUnload', client.onLogout)
+
+RegisterNetEvent('QBCore:Player:SetPlayerData', function(data)
+    if source == '' or not PlayerData.loaded then return end
+
+    if (data.metadata.isdead or data.metadata.inlaststand) ~= PlayerData.dead then
+        PlayerData.dead = data.metadata.isdead or data.metadata.inlaststand
+        OnPlayerData('dead', PlayerData.dead)
+    end
+
+    local groups = PlayerData.groups
+
+    if not groups[data.job.name] or not groups[data.gang.name] or groups[data.job.name] ~= data.job.grade.level or groups[data.gang.name] ~= data.gang.grade.level then
+        PlayerData.groups = {
+            [data.job.name] = data.job.grade.level,
+            [data.gang.name] = data.gang.grade.level,
+        }
+
+        OnPlayerData('groups', PlayerData.groups)
+    end
+
+    if data.metadata.ishandcuffed then
+        PlayerData.cuffed = true
+        LocalPlayer.state:set('invBusy', true, false)
+        Weapon.Disarm()
+    elseif PlayerData.cuffed then
+        PlayerData.cuffed = false
+        LocalPlayer.state:set('invBusy', false, false)
+    end
+end)
+
+---@diagnostic disable-next-line: duplicate-set-field
+function client.setPlayerStatus(values)
+    for name, value in pairs(values) do
+        -- compatibility for ESX style values
+        if value > 100 or value < -100 then
+            value = value * 0.0001
+        end
+
+        if name == "hunger" then
+            TriggerServerEvent('consumables:server:addHunger', QBCore.Functions.GetPlayerData().metadata.hunger + value)
+        elseif name == "thirst" then
+            TriggerServerEvent('consumables:server:addThirst', QBCore.Functions.GetPlayerData().metadata.thirst + value)
+        elseif name == "stress" then
+            if value > 0 then
+                TriggerServerEvent('hud:server:GainStress', value)
+            else
+                value = math.abs(value)
+                TriggerServerEvent('hud:server:RelieveStress', value)
+            end
+        end
+    end
+end
+
+AddStateBagChangeHandler('inv_busy', ('player:%s'):format(cache.serverId), function(_, _, value)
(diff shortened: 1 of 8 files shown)


JensGryspeert:
May I ask why closed? This is necessary for those with QBCore.

JensGryspeert:
Also I fixed some bugs you've introduced. Worth to check it out

HakkoDev:
> May I ask why closed? This is necessary for those with QBCore.

Qbcore support was removed for a reason, it will not be added back. If you want to use qb just use qbox, essentially the same thing but qbox doesn't have most of the cancerous growths qb does.

thelindat:
Qbus was dropped for many reasons including
- unauthorised use of proprietary code and assets
- violating licenses for open source resources
- general incompatibilities and breaking changes
- an inability to cooperate with and assist with the maintenance of resources they or many of their users rely on

Anytime something needed to be fixed, improved, or otherwise altered for the continued support of qbus it came to me to resolve it. Users of the infamously leaked non-compliant framework frequently expected flawless integration and failed to understand that none of their scripts were designed for nor in any way compatible with ox_inventory and other ox resources.

The amount of effort required to support the framework and provide technical support for its users was not in any way worth the time, nor the “reward” of being held responsible for their multitude of issues and incompetence.

There is simply no world that exists where their resources will be supported by, endorsed, or in any way recommended to any serious users intending to create a FiveM server.

Users of qbus are strongly encouraged to make the swap to qbox or literally any other framework, or otherwise maintain their own compatibility changes.

thelindat:
> Also I fixed some bugs you've introduced. Worth to check it out

There’s a reason we expect unrelated changes to be implemented in separate PRs.

unitysync:
img

PR #1957, opened 2026-06-12
fix(client): redefine data table by deepcloning it

LRemp:
**Issue:**
When the shops are loaded in the resource setup stage, the data is refferenced from the loaded config data table. When applying this operation repeatedly(for example changing player group data) the table data gets modified and leads into unwanted behavior - the shops become unavailable and in rare cases can lead into resource crashes.

**Fix:**
Deepclone the data table to prevent the data becoming malformed.

### DIFF
--- a/modules/shops/client.lua
+++ b/modules/shops/client.lua
@@ -172,7 +172,7 @@ local function refreshShops()
                                     distance = target.distance
                                 }
                             }),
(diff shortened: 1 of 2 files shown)


thelindat:
In what case is this table being modified by the player group?

PR #1959, opened 2026-06-17
Fixes the double restart after a client-side code update

NeenGame:
So when I restart ox_inventory and I have updated something in the client part, when it's cached, the inventory is printing return shared.info('cannot open inventory', '(player inventory has not loaded)') because PlayerData.loaded is false and I need to restart it again. But when I added the 250ms wait, it works fine with one restart.
I don't know if this is the correct way and if it's done on purpose — either way, this is why I made this pull request.

Screenshot 2026-06-18 015101


thelindat:
There's already a second delay before inventory setup, so adding an additional wait isn't really a fix so much as praying your computer loads fast enough.

PR #1964, opened 2026-07-12
feat(client): Add glovebox seat restriction

Cocodrulo:
## Summary
This pull request updates how the glovebox inventory can be accessed in vehicles, introducing a new configuration option to restrict glovebox access based on the player's seat. The changes ensure that glovebox access can be limited to certain seats, improving gameplay control and realism.

**Glovebox Access Restriction:**

* Added a new configuration option, `gloveboxseatrestriction`, which can be set via the `inventory:gloveboxseatrestriction` convar. When enabled, only players in the driver or front passenger seats (seat -1 or 0) can open the glovebox.
* Updated the logic in `client.lua` to check the new `gloveboxseatrestriction` setting before allowing glovebox access, ensuring that seat restrictions are enforced when configured. [[1]](diffhunk://#diff-bd3d5f0dfcec5f757b1cb2b6ec16fd75bb5ead988fed33033896aaf1a00e4a0dL784-R785) [[2]](diffhunk://#diff-bd3d5f0dfcec5f757b1cb2b6ec16fd75bb5ead988fed33033896aaf1a00e4a0dL823-R825)

<sup>\* Summary has been made by AI</sup>


thelindat:
While it’s named glovebox it’s really just any sort of internal storage space, I.e under or in the back of a seat, center console, etc. Hence why it’s typically available in other vehicles that aren’t cars. Further customisation is best handled using hooks.

Cocodrulo:
Well, It's true that further customization can be done with hooks, but there are two main reason why making it this way I think it would be better:
- The `cache.seat` and `cache.vehicle` is only accesible client-side.
- As it is called glovebox, most people will think that it's logical to be only accessed by the front passengers. It's a way to easily modify that restricttion without coding it in ox_inventory it self, without making a separate resource and with the possibility to disable the restriction (as it is by default) following the logic you mention about not only being glovebox.

By default it remains like it is, so it still works like now. Also with this addition we skip lots of code that it appears to be unnecessary if we want to make it work like a glovebox. If you still think it's a bad feature it's okey, but I think that adding this optional restriction is coherent and does not change the default behaviour.

iitzSamy:
I think it would be a good option for those who want this, if it's not enabled by default, only those who want that touch of realism would activate it

Cocodrulo:
what do you think @thelindat ?

Cocodrulo:
@thelindat

PR #1965, opened 2026-07-12
fix(inventory): avoid undefined bind parameter when saving empty inventory

oliann97:
## Summary
- Empty player inventory saved `nil` to oxmysql, which rejects `undefined` bind parameters with `"Bind parameters must not contain undefined. To pass SQL NULL specify JS null."`
- Replaced `next(buffer) and json.encode(buffer) or nil` with `json.encode(buffer)`, yielding `"[]"` for empty inventories
- Loading side treats `json.decode("[]")` (empty table) and SQL NULL identically — no behaviour change

## Reproduction
1. Player clears all inventory items
2. Trigger save (logout / autosave)
3. Console: `Error: ox_inventory was unable to execute a query!` + `Bind parameters must not contain undefined.`

## Changes
- `modules/inventory/server.lua` (`Inventory.Save`)
- `modules/inventory/server.lua` (`prepareInventorySave`)

## Notes
- Both call sites shared the same `next(buffer) and json.encode(buffer) or nil` pattern; both fixed
- No impact on stash/trunk/glovebox saves (they already pass `data` through unchanged)


thelindat:
Ensure you are using the correct version of oxmysql. Undefined parameters are already converted to null and that behaviour has only broken once for a short term release.

PR #1971, opened 2026-07-22
feat(server/inventory): add GetInventories export.

ad4ta:
Returns all loaded inventories, with an optional type filter.
There's no way to enumerate inventories from another resource, only single lookups via GetInventory.

Example:
```lua
-- clear every drop in the world
for _, drop in ipairs(exports.ox_inventory:GetInventories('drop')) do
    exports.ox_inventory:RemoveInventory(drop.id)
end


Bossman02:
I would make it so that if `invType` is `nil`, it directly returns `Inventories`; in that case, the `for` loop is completely unnecessary since all the inventories will be added to `returnData`.
Alternatively, I would set `returnData` to be equal to `Inventories`.

unitysync:
i don't really see any need/valid use case for this _at all_. 

nb: sending massive amounts of data across runtimes (lua -> node) is also not good practice and yields very little benefit.

ad4ta:
Resources that manage drops (death loot in my case) have to keep their own list of ids. That list goes stale because ox removes inventories on its own with the cleartime sweep. There is no way to ask "which drops exist right now."

Fair point on the data size. I can make invType required and remove items from the result, so it only returns id/type/label/owner/coords or just an id. Then you look up contents per id. Want me to change that?

Bossman02:
And add another parameter to specify whether you want the complete inventory lists or just the IDs? By default, it returns only the IDs.

ad4ta (commented):
Returns the ids of all loaded inventories of the given type, or the full inventories when detailed is set. Throws if invType is not a string.

thelindat:
Please create accompanying PRs for docs and types
https://github.com/overextended/overextended.github.io
https://github.com/overextended/ox_types/pulls

